PR #44, opened 2021-01-06
implementing validation

velimd:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/app/controller/AppRequest.ts
```diff
@@ -15,6 +14,6 @@ export interface AppRequest<T = Record<string, unknown>> extends Request {
 
 export interface AppSession extends Session {
   lang: string,
-  errors: DefinedError[] | undefined,
+  errors: [] | undefined,
```
linusnorton, line 17: what does `[]` even mean? array of what?

src/main/app/controller/GetController.ts
```diff
@@ -17,11 +17,11 @@ export class GetController {
     const commonLanguageContent = commonContent[req.session.lang] || commonContent['en'];
     const commonPageContent = this.content.common || {};
 
-    const errors = req.session.errors || [];
+    const sessionErrors = req.session.errors || [];
```
linusnorton, line 20: this seems a bit unnecessary now but okay

src/main/app/form/Form.ts
```diff
@@ -1,15 +1,25 @@
-import { DefinedError, ValidateFunction } from 'ajv';
 
 export class Form<T> {
 
   constructor(
-    private readonly validator: ValidateFunction
+    private readonly validator: any
```
linusnorton, line 5: There is a type for this... I think it's Form<T> or FormBody<T>

linusnorton: Also rename the property to form rather than validator

src/main/app/form/Form.ts
```diff
@@ -1,15 +1,25 @@
...
   ) { }
 
-  public getErrors(body: T): DefinedError[] {
-    this.validator(body);
-
-    return (this.validator.errors || []) as DefinedError[];
+  public getErrors(body: T): [] {
+    const errors = this.validator.fields
+      .filter(f => f.required)
```
linusnorton, line 10: hmmmmmmmmmmmmm. Rather than having a required property can you just check if there is a validator function?

src/main/steps/common/error/summary.njk
```diff
@@ -1,13 +1,13 @@
-{% if errors and errors.length > 0 %}
+{% if sessionErrors and sessionErrors.length > 0 %}
```
linusnorton, line 1: you sure about this rename?

src/main/steps/screen-questions/language-preference/content.ts
```diff
@@ -13,18 +20,25 @@ const cy: typeof en = {
...
+      required: 'Dewisiwch yr iaith yr hoffech gael eich dogfennau a’ch gohebiaeth.'
+    }
+  }
 };
 
 export const languagePreferenceForm = {
-  fields: {
-    languagePreferenceWelsh: {
+  fields: [
+    { name: 'languagePreferenceWelsh',
```
linusnorton, line 32: I would prefer to use the key name, any reason you didnt?

src/main/steps/screen-questions/language-preference/content.ts
```diff
@@ -13,18 +20,25 @@ const cy: typeof en = {
...
       type: 'radios',
       values: [
         { label: l => l.no, value: 'No' },
         { label: l => l.yes, value: 'Yes' }
-      ]
+      ],
+      required: true,
```
linusnorton, line 38: as mentioned above, I think this probably isn't necessary as you can just use the presence of the validator to say whether it's required or not

src/main/utils/validation.ts
```diff
@@ -0,0 +1,4 @@
+
+export const isFieldFilledIn = (value: string): (boolean | string)[] => {
+  return [!!value, 'required'];
```
linusnorton, line 3: I don't think this will work the way I imagined.

I think it should return false if there are no errors or the error type if there is an error. That allows you to chain together multiple checks:

```
validation: field => isFieldFilledIn(field) || isFieldLongerThan(field, 3) || isFieldShorterThan(field, 255)
```

The first one that returns something truthy (e.g. 'required') would be the return value and be the error.

src/main/app/controller/AppRequest.ts
velimd: ill make a type, plan was to have { propertyName, errorType }

src/main/app/form/Form.ts
linusnorton: maybe content

src/main/app/controller/GetController.ts
velimd: its because in content, errors already exist so i thought ill change it here, if we are going to copy the content over, itll reduce mistakes but we can change it in the content as well.

src/main/app/form/Form.ts
velimd: yh, i forgot

src/main/app/form/Form.ts
velimd: ill have a look

src/main/steps/screen-questions/language-preference/content.ts
velimd: maybe its coz im rusty but i couldnt do any filtering or map functions to property, only thing could think of was to covert it into an array object but if you know any othe way im game.

linusnorton:
```
Object.values(form.fields).map(...)
```

src/main/app/controller/GetController.ts
linusnorton: I understand

linusnorton (approved):

linusnorton (approved):

PR #59, opened 2021-01-13
Add language toggle

linusnorton:
### JIRA link ###

https://tools.hmcts.net/jira/browse/NFDIV-298

### Change description ###

Add language toggle and beta banner. Remove sass-lint as it's no longer maintained.

### DIFF
--- a/src/main/modules/session/index.ts
+++ b/src/main/modules/session/index.ts
@@ -1,4 +1,4 @@
-import { Application } from 'express';
+import { Application, Request } from 'express';
 import session from 'express-session';
 import ConnectRedis from 'connect-redis';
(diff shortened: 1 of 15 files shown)


src/main/modules/session/index.ts
```diff
@@ -48,3 +49,7 @@ export class SessionStorage {
       });
   }
 }
+
+export type ReqWithSession = Request & {
```
velimd, line 53: i added something in my pr for this, tho you way seems cleanier

`declare module 'express-session' {
  export interface SessionData {
    user: Record<string, unknown>
  }
}
`

velimd (approved):

PR #70, opened 2021-01-15
FOR JON, ALWAYS

velimd:
### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/modules/oidc/index.ts
+++ b/src/main/modules/oidc/index.ts
@@ -41,7 +41,10 @@ export class OidcMiddleware {
 
     server.get('/logout', function(req: Request, res){
       req.session.user = undefined;
-      req.session.save(() => res.redirect('/'));
+      req.session.save(() => {
+        res.redirect('/');
+        res.locals.isLoggedIn = false;
+      });
     });
 
     server.use((req: Request, res: Response, next: NextFunction) => {
--- a/src/main/modules/oidc/index.ts
+++ b/src/main/modules/oidc/index.ts
@@ -41,18 +41,15 @@ export class OidcMiddleware {
 
     server.get('/logout', function(req: Request, res){
       req.session.user = undefined;
-      req.session.save(() => {
-        res.redirect('/');
-        res.locals.isLoggedIn = false;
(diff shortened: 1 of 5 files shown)


src/main/modules/oidc/index.ts
```diff
@@ -41,7 +41,10 @@ export class OidcMiddleware {
 
     server.get('/logout', function(req: Request, res){
       req.session.user = undefined;
-      req.session.save(() => res.redirect('/'));
+      req.session.save(() => {
+        res.redirect('/');
+        res.locals.isLoggedIn = false;
```
linusnorton, line 46: do you mean to set the session?

velimd: na, i think the placement is wrong hence ill change it, but in new version i pushed, in my opinion every request is going to hit that callback so it will check it there.

`if (req.session.user) {
        res.locals.isLoggedIn = true;
        return next();
      }
      res.locals.isLoggedIn = false;
      res.redirect('/login');`

velimd: otherwise we'll have to return if their authenticated in each render response. i think this way is cleaner.

SarahLittlejohn (approved):

src/main/modules/oidc/index.ts
```diff
@@ -47,9 +47,9 @@ export class OidcMiddleware {
     server.use((req: Request, res: Response, next: NextFunction) => {
       if (req.session.user) {
         res.locals.isLoggedIn = true;
-
         return next();
       }
+      res.locals.isLoggedIn = false;
```
linusnorton, line 52: why set a local if you are redirecting? do you mean to set the session?

linusnorton (approved):

src/main/modules/oidc/index.ts
velimd: i can have a session value called loggedin but i dont see the point (right now) if we have the user one. I can assign it before the if statement so it checks if the user object is there but again, i dont see the point as res.locals only is available during the request.

PR #84, opened 2021-01-20
Chore/fix eslint warnings

velimd:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/app/controller/GetController.ts
```diff
@@ -8,7 +8,7 @@ export class GetController {
 
   constructor(
     protected readonly name: string,
-    protected readonly content: Record<string, any>
+    protected readonly content: Record<string, Record<string, unknown>>
```
velimd, line 11: i dont think this is true, content object is mostly property and string value with the expection of erros hence i see why its any but i dont think its an object with values being objects.

src/main/app/form/Form.ts
```diff
@@ -14,7 +14,7 @@ export class Form {
       .filter((key) => this.form.fields[key].validator !== undefined)
       .reduce((errors: FormError[], propertyName: string) => {
         const field = this.form.fields[propertyName];
-        const errorType = field.validator!(body[propertyName]);
+        const errorType = field.validator?.(body[propertyName] as string);
```
velimd, line 17: so is the ? the same as ! operator?

src/main/modules/i18n/index.ts
```diff
@@ -8,12 +7,12 @@ export class LanguageToggle {
   private supportedLanguages = ['en', 'cy'];
 
   public enableFor(app: Express): void {
-    app.use((req: ReqWithSession, res: Response, next: NextFunction) => {
+    app.use((req, res, next) => {
```
velimd, line 10: dont we define types in typescript, wouldnt it be best to have its type?

src/main/modules/i18n/index.ts
```diff
@@ -8,12 +7,12 @@ export class LanguageToggle {
...
       if (req.method === 'GET' && req.query['lng']) {
         const requestedLanguage = req.query['lng'] as string;
 
         if (this.supportedLanguages.includes(requestedLanguage)) {
-          req.session.lang = requestedLanguage;
+          req.session['lang'] = requestedLanguage;
```
velimd, line 15: probably related to removing types or is this a preferred way?

src/main/server.ts
```diff
@@ -44,7 +44,7 @@ new AppInsights().enable();
 new SessionStorage().enableFor(app);
 new CSRFToken().enableFor(app);
 new LanguageToggle().enableFor(app);
-new OidcMiddleware().enableFor(app);
+// new OidcMiddleware().enableFor(app);
```
velimd, line 47: is there a reason we are commenting these out?

sebflipper: My bad, didn't mean to commit this file, since that's only need to get the app working.

src/main/app/form/Form.ts
sebflipper: Not quite:

* `!.` [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) (TypeScript only). It's a way to tell the compiler that "this expression cannot be `null` or `undefined` here" (so don't show any errors)
* `?.` [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) (works in both TS & JS). Returns `undefined` if the property or method doesn't exist

src/main/modules/i18n/index.ts
sebflipper: I was getting an error here since the line above is `app: Express` which provides the types the `app.use()` method and by retyping the function call here, I was getting an error saying that the retyped arguments are incompatible the existing `app.use()` type.

src/main/app/form/Form.ts
velimd: thanks for the explainations, in that case, shouldnt ! still be used as we can filtering that field and checking its not undefined before hand. I guess its fine either way, so im good either way.

src/main/app/controller/GetController.ts
sebflipper: I double checked the usage of `content` and it does seem to be a language object of objects such as:

```js
const en = {
  question: 'Has your marriage broken down irretrievably (it can’t be saved)?',
  line1: 'You can only get a divorce on the ground that your marriage has broken down irretrievably.',
  yes: 'Yes',
  no: 'No',
  marriageBrokenDownNo: 'The only legal ground for divorce in England and Wales is that your marriage has broken down irretrievably.',
  errors: {
    'screenHasMarriageBroken': {
      'required': 'Select yes if your marriage has broken down irretrievably'
    }
  }
};
```

src/main/modules/i18n/index.ts
sebflipper: `req.query` is used in the same way above (`req.query['lng']`). So `req.session['lang']` is used to match the type from `app: Express`.

src/main/app/controller/GetController.ts
velimd: oh yh, my bad, i was thinking of the en object itself.

linusnorton: Rather than use `Record<string, unknown>` I think we have `AnyObject` but that might be `Record<string, any>` I can't remember

src/main/app/form/Form.ts
linusnorton: IMO I think ! is fine because of the filter but I'm happy either way

src/main/modules/helmet/index.ts
```diff
@@ -7,7 +7,7 @@ export interface HelmetConfig {
 }
 
 const googleAnalyticsDomain = '*.google-analytics.com';
-const self = "'self'";
+const self = '\'self\'';
```
linusnorton, line 10: don't be that guy

src/main/steps/screen-questions/language-preference/content.ts
```diff
@@ -27,15 +27,15 @@ const cy: typeof en = {
...
+export const languagePreferenceForm: FormContent = {
   fields: {
     languagePreferenceWelsh: {
       type: 'radios',
       values: [
         {label: l => l.no, value: 'No'},
         {label: l => l.yes, value: 'Yes'}
       ],
-      validator: value => isFieldFilledIn(value)
+      validator: (value: string): string | void => isFieldFilledIn(value)
```
linusnorton, line 38: this is unfortunate - it's quite long winded now

linusnorton (approved):
few comments but mostly fine

src/main/modules/helmet/index.ts
sebflipper: Must have been my IDE 😅 I can revert this file

src/main/app/form/Form.ts
sebflipper: Our current eslint rules from the TypeScript recommended ruleset, mark this line as a `Forbidden non-null assertion. eslint(@typescript-eslint/no-non-null-assertion)`.

Possible because if you're saying that the `field.validator` field might not exist, but then trying to use it by force. If it is actually optional then you'll get an error trying to invoke a non-existent function. Using `?.` would prevent this error.

However, maybe a better option would be to mark this field as required. I'll take a look if that could work.

linusnorton: no I'm happy with ?

src/main/steps/screen-questions/language-preference/content.ts
sebflipper: This might work better if I made it into a reusable `Validator` type.

PR #85, opened 2021-01-20
chore: add Prettier ESLint config

velimd:
### JIRA link ###

N/A

### Change description ###

This PR introduces [Prettier code formatting](https://prettier.io/), which:

* Automatically formats code on save and commit
* No need to discuss style in code review
* Saves you time and energy
* [And more](https://prettier.io/docs/en/why-prettier.html)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/modules/appinsights/index.ts
+++ b/src/main/modules/appinsights/index.ts
@@ -3,16 +3,12 @@ import config from 'config';
 const appInsights = require('applicationinsights');
 
 export class AppInsights {
-
   enable(): void {
     if (config.get('appInsights.instrumentationKey')) {
-      appInsights.setup(config.get('appInsights.instrumentationKey'))
-        .setSendLiveMetrics(true)
-        .start();
+      appInsights.setup(config.get('appInsights.instrumentationKey')).setSendLiveMetrics(true).start();
 
       appInsights.defaultClient.context.tags[appInsights.defaultClient.context.keys.cloudRole] = 'nfdiv-frontend';
-      appInsights.defaultClient.trackTrace({message: 'App insights activated'});
+      appInsights.defaultClient.trackTrace({ message: 'App insights activated' });
     }
   }
-
 }
--- a/src/main/steps/common/common.content.ts
+++ b/src/main/steps/common/common.content.ts
@@ -1,32 +1,37 @@
 const en = {
   phase: 'Beta',
-  feedback: 'This is a new service – your <a class="govuk-link" aria-label="Feedback link, This will open a new tab. You’ll need to return to this tab and continue with your application within 60 mins so you don’t lose your progress." href="https://www.smartsurvey.co.uk/s/Divorce_Feedback" target="_blank">feedback</a> will help us to improve it.',
+  feedback:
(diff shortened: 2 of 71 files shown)


src/main/steps/common/common.content.ts
```diff
@@ -1,32 +1,37 @@
 const en = {
   phase: 'Beta',
-  feedback: 'This is a new service – your <a class="govuk-link" aria-label="Feedback link, This will open a new tab. You’ll need to return to this tab and continue with your application within 60 mins so you don’t lose your progress." href="https://www.smartsurvey.co.uk/s/Divorce_Feedback" target="_blank">feedback</a> will help us to improve it.',
+  feedback:
+    'This is a new service – your <a class="govuk-link" aria-label="Feedback link, This will open a new tab. You’ll need to return to this tab and continue with your application within 60 mins so you don’t lose your progress." href="https://www.smartsurvey.co.uk/s/Divorce_Feedback" target="_blank">feedback</a> will help us to improve it.',
```
velimd, line 4: dont really see the point of this being on a new line as it will still be quite long.

src/test/functional/utlis/puppeteer.util.ts
```diff
@@ -19,15 +21,15 @@ export const goTo = async (url: string) => {
 };
 
 export const getPageTitle = async () => {
-  return await scope.page.title();
+  return scope.page.title();
```
velimd, line 24: i think the await is quite important in puppeteer calls. not sure if it needed here tho.

src/main/modules/appinsights/index.ts
```diff
@@ -3,16 +3,12 @@ import config from 'config';
 const appInsights = require('applicationinsights');
 
 export class AppInsights {
-
   enable(): void {
     if (config.get('appInsights.instrumentationKey')) {
-      appInsights.setup(config.get('appInsights.instrumentationKey'))
-        .setSendLiveMetrics(true)
-        .start();
+      appInsights.setup(config.get('appInsights.instrumentationKey')).setSendLiveMetrics(true).start();
```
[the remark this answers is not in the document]
sebflipper: The Prettier recommended default for `printWidth` is `80`.

This can be configured to be [shorter](https://prettier.io/docs/en/options.html#print-width), which would wrap these lines, but then would also wrap other parts of the code.

src/main/steps/common/common.content.ts
sebflipper: Prettier by default is configured to wrap anything greater than a line length of `80`, this can be [configured](https://prettier.io/docs/en/options.html#print-width), but then would go against the wrapping you mentioned above.

src/test/functional/utlis/puppeteer.util.ts
sebflipper: The `await` here is [unnecessary](https://eslint.org/docs/rules/no-return-await) since the whole function is an `async` method, so the caller will need to await it again: `await getPageTitle()`.

sebflipper:
@velimd I'm closing this PR, since you're unable to approve my branch as you created it and I'm going to recreate this PR from my account 👍

PR #107, opened 2021-01-25
feat(NFDIV-379): handle 4xx, 5xx errors, timeouts, and asynchronous errors 

sebflipper:
### JIRA link ###

[NFDIV-379](https://tools.hmcts.net/jira/browse/NFDIV-379) Handle 400 & 500 errors

### Change description ###

* Adds English and Welsh translations for `BAD_REQUEST`, `NOT_FOUND`, `INTERNAL_SERVER_ERROR` similar to [hmcts/div-petitioner-frontend](https://github.com/hmcts/div-petitioner-frontend)
* This sets the max time for incoming and outgoing requests to 30 seconds
* Shows an error if any controller throws an error or doesn't handle a rejected promise e.g.
```ts
public async get(req: AppRequest, res: Response, next: NextFunction): Promise<void> {
  throw new Error('foo'); // uncaught error
  await new Promise((res, rej) => setTimeout(rej, 1000)); // unhandled promise rejection
  await new Promise((res, rej) => setTimeout(res, 31000)); // timeout after 30 seconds
  return next('error'); // returning a `next` Express error

  next('error'); // calling an Express `next` with an error and rendering afterwards
  super.get(req, res);
}
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "glob": "^7.1.6",
     "govuk-frontend": "^3.10.2",
     "helmet": "^4.4.1",
+    "http-status-codes": "^2.1.4",
     "jwt-decode": "^3.1.2",
     "lodash": "^4.17.20",
     "nunjucks": "^3.2.2",
--- a/src/main/app/controller/GetController.ts
+++ b/src/main/app/controller/GetController.ts
@@ -12,7 +12,7 @@ export class GetController {
   ) {}
 
   public async get(req: AppRequest, res: Response): Promise<void> {
-    setImmediate(() => {
+    setImmediate(async () => {
       if (res.headersSent) {
         // If there's an async error, it may have already rendered an error page upstream,
         // so we don't want to call render again if headers have already been sent
--- a/src/main/app/controller/GetController.ts
+++ b/src/main/app/controller/GetController.ts
@@ -12,22 +12,20 @@ export class GetController {
   ) {}
 
   public async get(req: AppRequest, res: Response): Promise<void> {
-    setImmediate(async () => {
-      if (res.headersSent) {
-        // If there's an async error, it may have already rendered an error page upstream,
-        // so we don't want to call render again if headers have already been sent
-        return;
-      }
+    if (res.locals.isError) {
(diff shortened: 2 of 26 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
     "glob": "^7.1.6",
     "govuk-frontend": "^3.10.2",
     "helmet": "^4.4.1",
+    "http-status-codes": "^2.1.4",
```
linusnorton, line 61: oh JavaScipt... of course there is a package:)

src/main/app/controller/GetController.ts
```diff
@@ -12,15 +12,22 @@ export class GetController {
...
   public async get(req: AppRequest, res: Response): Promise<void> {
-    const languageContent = this.content[req.session.lang] || this.content['en'] || {};
-    const commonLanguageContent = commonContent[req.session.lang] || commonContent['en'];
-    const commonPageContent = this.content.common || {};
+    setImmediate(async () => {
+      if (res.headersSent) {
+        // If there's an async error, it may have already rendered an error page upstream,
+        // so we don't want to call render again if headers have already been sent
+        return;
+      }
```
linusnorton, line 20: I'm sorry, I don't think this is a viable solution. I'm not willing to wrap all controllers in a setImmediate. We can switch to express 5.x alpha if we have to, but there's got to be a better way

linusnorton (changes requested):
I'm okay with everything except the setImmediate

package.json
sebflipper: Was "borrowed" from [hmcts/div-petitioner-frontend](https://github.com/hmcts/div-petitioner-frontend) since it reads better than status code numbers 😄

src/main/app/controller/GetController.ts
sebflipper: 👍 I've removed `setImmediate` in https://github.com/hmcts/nfdiv-frontend/pull/107/commits/719bddc3bb808d05519972616381ea2609e9fc98

linusnorton (approved):

PR #125, opened 2021-01-27
Nfdiv 385/case created

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-385

### Change description ###
- call idam to get users details as we arent aware of it, we could take it from the jwt_decode, might be risky
- call cos api to create case as the user logs in.

Note: i havent included test, just curious on the approach.

```
[ ] Yes
[X] No
```


src/main/modules/api/cos-api.ts
```diff
@@ -0,0 +1,19 @@
+import { AxiosInstance } from 'axios';
+import { LoggerInstance } from 'winston';
+
+export class COSApi {
```
linusnorton, line 4: I think the filename and class name should be CosApi. It's the convention I normally but you don't have to

src/main/modules/api/cos-api.ts
```diff
@@ -0,0 +1,19 @@
...
+  constructor(
+    private readonly axios: AxiosInstance,
+    private readonly logger: LoggerInstance
+  ) { }
+
+  public createCase(data: any): Promise<boolean> {
```
linusnorton, line 10: Maybe something better than any?

src/main/modules/api/cos-api.ts
```diff
@@ -0,0 +1,19 @@
...
+    return this.axios
+      .post('/submit', data)
+      .then(results => !!results)
```
linusnorton, line 13: I assume this is just temporary?

src/main/modules/oidc/index.ts
```diff
@@ -47,8 +49,24 @@ export class OidcMiddleware {
...
-    app.use((req: Request, res: Response, next: NextFunction) => {
+    app.use(async (req: Request, res: Response, next: NextFunction) => {
       if (req.session?.user) {
+        const user = req.session.user;
+        req.scope = req.app.locals.container.createScope();
+        req.scope?.register({
+          axios: asValue(Axios.create({
+            baseURL: config.get('services.cos.baseURL'),
+            headers: {
+              Authorization: 'Bearer ' + user.id_token
```
linusnorton, line 60:
```suggestion
              Authorization: 'Bearer ' + user.access_token,
              IdToken: user.id_token
```

src/main/modules/oidc/index.ts
```diff
@@ -47,8 +49,24 @@ export class OidcMiddleware {
...
+            }
+          })),
+          api: asClass(COSApi)
+        });
+
+        if (!req.session.caseCreated) {
```
linusnorton, line 66: do we want to try a GET /case before creating one?

src/main/modules/oidc/index.ts
```diff
@@ -59,6 +77,13 @@ export class OidcMiddleware {
 
 declare module 'express-session' {
   export interface SessionData {
-    user: Record<string, unknown>
+    user: Record<string, Record<string, unknown>>,
+    caseCreated: boolean
+  }
+}
+
+declare module 'Express' {
```
linusnorton, line 85: is there another way of doing this?

src/main/modules/api/cos-api.ts
```diff
@@ -0,0 +1,19 @@
...
+      .catch(err => {
+        this.logger.error(err);
+        throw new Error('Case could not be created. Please try again.');
+      });
```
sebflipper, line 17: You could make this easier to read by making it async e.g.

```suggestion
      let results = false;
      try {
        results = !!(await this.axios.post('/submit', data));
      } catch (e) {
        this.logger.error(err);
        throw new Error('Case could not be created. Please try again.');
      }
      return results;
```

src/main/modules/api/cos-api.ts
velimd: initially i wrote that but i thought as COS is an acronym, it might be best to name it like that but i understand. Ill change it.

src/main/modules/api/cos-api.ts
velimd: oh, initially i wasnt sure what the payload would be but ill change it after talking to Sarah.

src/main/modules/api/cos-api.ts
velimd: Again, wasnt sure what is being returned so i left it as, if theres data, there is a case.

src/main/modules/oidc/index.ts
velimd: I wasnt sure the endpoint was created yet so i left it like that, ill speak to sarah to see.

src/main/modules/api/cos-api.ts
velimd: imo i think the then and catch make is as easier to read plus looks nicer visually. Will discuss with sarah what is being returned so at that point, there might be difficulty there so the approach might be better but even then i think the then block can handle it.

linusnorton (approved):

PR #147, opened 2021-02-03
clean up

velimd:
### Change description ###
clean up

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/test/a11y/a11y.ts
+++ b/src/test/a11y/a11y.ts
@@ -20,6 +20,22 @@ interface PallyIssue {
   typeCode: number;
 }
 
+function loginPally(): Pa11yResult {
+  return pa11y(`${config.TEST_URL}/login`, {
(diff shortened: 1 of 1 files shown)


src/test/a11y/a11y.ts
```diff
@@ -20,19 +27,29 @@ interface PallyIssue {
...
+function loginPally(): Pa11yResult {
+  return pa11y(`${config.TEST_URL}/login`, {
+    hideElements: '.govuk-footer__licence-logo, .govuk-header__logotype-crown',
+    actions: [
+      `set field #username to ${config.TestUser}`,
+      `set field #password to ${config.TestPass}`,
+      'click element .button',
+      'wait for path to be /',
+    ],
+  });
```
SebFlippence, line 39: Looks like each call of `pa11y()` spins up it's own browser so I don't think you can reuse instances 😞

PR #151, opened 2021-02-04
Certificate path

velimd:
Note: This is just a draft to include stuff such as civil content or addition page if 'no' was selected.


src/main/steps/exit/no-certificate/get.ts
```diff
@@ -0,0 +1,9 @@
+import { GetController } from '../../../app/controller/GetController';
+
+import { noCertificateContent } from './content';
+
+export class noCertificateGetController extends GetController {
```
linusnorton, line 5: NoCertificateGetController

PR #156, opened 2021-02-05
Nfdiv 385/case created

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-385

### Change description ###

- call idam to get users details as we arent aware of it, we could take it from the jwt_decode, might be risky
- call cos api to create case as the user logs in.

Note: i havent included test, just curious on the approach.

```
[ ] Yes
[X] No
```


charts/nfdiv-frontend/values.yaml
```diff
@@ -16,9 +16,10 @@ nodejs:
         - AppInsightsInstrumentationKey
         - redis-access-key
   environment:
-    IDAM_WEB_URL: "https://idam-web-public.{{ .Values.global.environment }}.platform.hmcts.net/login"
-    IDAM_API_URL: "https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net/o/token"
-    REDIS_HOST: "nfdiv-frontend-redis-{{ .Values.global.environment }}.redis.cache.windows.net"
+    IDAM_WEB_URL: 'https://idam-web-public.{{ .Values.global.environment }}.platform.hmcts.net/login'
+    IDAM_API_URL: 'https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net/o/token'
+    REDIS_HOST: 'nfdiv-frontend-redis-{{ .Values.global.environment }}.redis.cache.windows.net'
+    COS_URL: 'http://nfdiv-cos-{{ .Values.global.environment }}.service.core-compute-aat.internal'
```
nitinprabhu, line 22: you will also need to update `values.preview.template.yaml` for `COS_URL` to have value `http://nfdiv-cos-aat.service.core-compute-aat.internal` assuming in preview you want to use AAT ?

src/main/modules/oidc/index.ts
```diff
@@ -57,8 +60,29 @@ export class OidcMiddleware {
...
       if (req.session?.user) {
+        const user = req.session.user;
+        req.scope = req.app.locals.container.createScope();
+        req.scope?.register({
+          axios: asValue(
+            Axios.create({
+              baseURL: config.get('services.cos.baseURL'),
+              headers: {
+                Authorization: 'Bearer ' + user.access_token,
+                ServiceAuthorization: user.id_token,
```
linusnorton, line 73: not ServiceAuthorization, it's IdToken

linusnorton, line 73: #
```suggestion
                IdToken: user.id_token,
```

linusnorton (approved):
Okay, two things that we can save for later:

- We'll need a definition on that response from the API.
- We should move the case creation logic to the index controller and then tie it in with the logic to work out what page to go back to

PR #174, opened 2021-02-10
Nfdiv 385/case created

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-452

### Change description ###
Call to create case when a user logs in.

```
[ ] Yes
[X] No
```


src/main/modules/api/cos-api.ts
```diff
@@ -0,0 +1,34 @@
+import { AxiosInstance } from 'axios';
+import { LoggerInstance } from 'winston';
+
+export class CosApi {
```
SebFlippence, line 4: Perhaps we should start adding unit tests for new modules?

linusnorton: this shouldn't be a module and it shouldn't be cos-api.ts :)

linusnorton: modules should just be for express middleware setup

src/main/app/api/CosApi.ts
```diff
@@ -0,0 +1,34 @@
...
+  //TODO change return type when backend has been implemented.
+  public getCase(): Promise<Record<string, unknown>> {
+    return this.axios
+      .get('/retrieve-case')
+      .then(results => results.data)
+      .catch(err => {
+        this.logger.error(err);
+        return false;
+      });
+  }
```
SebFlippence, line 16: Looks like this method needs a test to get it up [above the 60% coverage from sonarcloud](https://sonarcloud.io/component_measures?id=nfdiv-frontend&pullRequest=174&metric=new_coverage&view=list)

config/default.yaml
```diff
@@ -8,6 +8,8 @@ services:
     tokenURL: 'https://idam-api.aat.platform.hmcts.net/o/token'
     clientID: 'divorce'
     clientSecret: 'NEED TO INSERT SECRET'
+  cos:
+    baseURL: 'http://localhost:4012'
```
SebFlippence, line 12: Do we need to add some instructions to the readme about how to get this service running locally?

velimd: included this however, you just need to run cos, nothing special config needed, let me know if i need to include more.

SebFlippence (approved):

charts/nfdiv-frontend/values.preview.template.yaml
```diff
@@ -8,6 +7,9 @@ nodejs:
     prefix: civil
     primaryIngressHost: ${SERVICE_FQDN}
   environment:
+    IDAM_WEB_URL: 'https://idam-web-public.aat.platform.hmcts.net/login'
```
linusnorton, line 10: you don't need these three lines

linusnorton (approved):

SebFlippence (approved):

PR #175, opened 2021-02-10
feat(NFDIV-440): Added Save & sign out link with session storage update

SebFlippence:
### JIRA link ###

Part of [NFDIV-440](https://tools.hmcts.net/jira/browse/NFDIV-440) English Version - Who are you applying to divorce / end a civil partnership with

### Change description ###

* Adds the "Save and sign out" link at the bottom of the form, which triggers a form submit, saving the current state without validation
* Sends the current form data session state to the current step/page
* Fixes an issue where the session state was not always saving
* Passes the current form state back into the form to restore the users input

```
[ ] Yes
[x] No
```


src/main/app/controller/GetController.ts
```diff
@@ -10,7 +10,11 @@ export type TranslationFn = (isDivorce: boolean) => Translations;
 
 @autobind
 export class GetController {
-  constructor(protected readonly view: string, protected readonly content: TranslationFn | Translations) {}
+  constructor(
+    protected readonly view: string,
+    protected readonly content: TranslationFn | Translations,
+    protected readonly id: string = 'page'
```
velimd, line 16: how comes you are setting a default value here?

src/main/modules/nunjucks/index.ts
```diff
@@ -19,11 +19,11 @@ export class Nunjucks {
       return typeof prop === 'function' ? prop(this.ctx) : prop;
     });
 
-    env.addGlobal('formItems', function (items: FormInput[]) {
+    env.addGlobal('formItems', function (items: FormInput[], selected: string) {
       return items.map(i => ({
         text: this.env.globals.getContent.call(this, i.label),
         value: i.value,
-        selected: i.selected,
+        checked: i.selected || i.value === selected,
```
velimd, line 26: im confused on i.selected, would a form ever have a default value other than the one the user selected before and saved in session?

velimd: i guess its to be generic component.

src/main/app/controller/GetController.ts
SebFlippence: 👍 Good spot, I've renamed this to `stepId` to make it more obvious what it's referring to and set it to undefined by default for pages that don't need a form in bb0bf46.

src/main/modules/nunjucks/index.ts
SebFlippence: Yeah, this is for generic input options. `i.selected` is from the `form` object in `content.ts`. So you could always make sure that an option is always selected by default. The second part is the users selection so I've renamed this to be `userSelection` in b07abed 👍

velimd (approved):

PR #179, opened 2021-02-11
feat(NFDIV-440): Send spouse to content + clear session on sign out

SebFlippence:
### JIRA link ###

Part of [NFDIV-440](https://tools.hmcts.net/jira/browse/NFDIV-440) English Version - Who are you applying to divorce / end a civil partnership with

### Change description ###

* Clears session details on sign out and set cookie timeout to 20 minutes
* Sends the spouse/partner type onto the current content e.g.

```ts
export const generateContent = (): TranslationFn => ({ isDivorce, partner })
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/controller/GetController.ts
+++ b/src/main/app/controller/GetController.ts
@@ -6,7 +6,7 @@ import { commonContent } from '../../steps/common/common.content';
 import { AppRequest } from './AppRequest';
 
 export type Translations = Record<'en' | 'cy' | 'common', Record<string, unknown>>;
-export type TranslationFn = (isDivorce: boolean) => Translations;
+export type TranslationFn = ({ isDivorce, partner }: { isDivorce: boolean; partner: string }) => Translations;
 
 @autobind
 export class GetController {
@@ -23,13 +23,11 @@ export class GetController {
       return;
     }
 
-    const isDivorce = res.locals.serviceType !== 'civil';
-    const derivedContent = typeof this.content === 'function' ? this.content(isDivorce) : this.content;
-
     const language = req.session.lang || 'en';
-    const languageContent = derivedContent[language];
     const commonLanguageContent = commonContent[language];
-    const commonPageContent = derivedContent.common || {};
+    const content = this.getContent(req, res, commonLanguageContent);
+    const languageContent = content[language];
+    const commonPageContent = content.common || {};
 
     const sessionErrors = req.session.errors || [];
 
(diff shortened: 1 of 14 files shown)


src/main/app/controller/GetController.ts
```diff
@@ -40,7 +38,27 @@ export class GetController {
...
+
+  private getContent(req: AppRequest, res: Response, translations: Translations): Translations {
+    if (typeof this.content !== 'function') {
+      return this.content;
+    }
+
+    const isDivorce = res.locals.serviceType !== 'civil';
+    const yourDetails = req.session.state?.['your-details'] as Record<string, string> | undefined;
+    const selectedPartnerGender = yourDetails?.partnerGender;
+    let partner = '';
```
linusnorton, line 53: rather than  a let maybe you could split out getPartner into a method with return values

src/main/modules/oidc/index.ts
```diff
@@ -55,9 +56,11 @@ export class OidcMiddleware {
 
     app.get(
       SIGN_OUT_URL,
-      errorHandler((req: Request, res) => {
-        req.session.user = undefined;
-        req.session.userCase = undefined;
+      errorHandler((req: AppRequest, res) => {
+        delete req.session.user;
```
linusnorton, line 60: I minor detail but I usually set things to undefined rather than using delete. The delete keyword re-indexes things behind the scenes and gets a bit slow on large objects/collections.

linusnorton (approved):

src/main/app/controller/GetController.ts
SebFlippence: 👍 Good idea, I've added this in cbd0408.

src/main/modules/oidc/index.ts
SebFlippence: 👍 I've changed these back in cbd0408

src/main/app/controller/GetController.ts
```diff
@@ -40,7 +38,27 @@ export class GetController {
...
     });
   }
+
+  private getContent(req: AppRequest, res: Response, translations: Translations): Translations {
+    if (typeof this.content !== 'function') {
+      return this.content;
+    }
+
+    const isDivorce = res.locals.serviceType !== 'civil';
+    const yourDetails = req.session.state?.['your-details'] as Record<string, string> | undefined;
```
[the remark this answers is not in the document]
SebFlippence: 👍 Looks like the typing was off for this property, so I've fixed it so we don't need to split and cast into 2 variables

PR #191, opened 2021-02-15
Nfdiv 378/date formed

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-378

### Change description ###
Adding the date formed page,
included a parser option in form fields as the date is split up into 3 different fields (thanks to gds) therefore having a parser which changes the 3 date objects into 1 object.

```
[ ] Yes
[X] No
```


src/main/steps/sequence.test.ts
```diff
@@ -30,7 +35,7 @@ describe('Sequence', () => {
 
     it('returns current step if there is no next step', () => {
       mockReq.originalUrl = HAS_RELATIONSHIP_BROKEN_URL;
-      expect(getNextStepUrl(mockReq)).toBe(HAS_RELATIONSHIP_BROKEN_URL);
+      expect(getNextStepUrl(mockReq)).toBe(RELATIONSHIP_DATE_URL);
```
SebFlippence, line 38: I think this test should be:

```ts
      const lastUrl = RELATIONSHIP_DATE_URL;
      mockReq.originalUrl = lastUrl;
      expect(getNextStepUrl(mockReq)).toBe(lastUrl);
```

velimd: yes but i think original should be the url before the last url. I assume youre going for that idea
`const lastUrl = RELATIONSHIP_DATE_URL;
      mockReq.originalUrl = HAS_RELATIONSHIP_BROKEN_URL;
      expect(getNextStepUrl(mockReq)).toBe(lastUrl);`

src/main/steps/sequence/relationship-date/template.njk
```diff
@@ -0,0 +1,6 @@
+{% extends "common/page.njk" %}
+
+{% block page_content %}
+  <h1 class="govuk-heading-l">{{ title }}</h1>
+  <p class="govuk-body">{{ line1 }}</p>
+{% endblock %}
```
SebFlippence, line 6: You should be able to delete this file since it will use the the default `src/main/steps/sequence/template.njk` (which is the same) if there's no template file.

SebFlippence (approved):

src/test/functional/features/marriage-broken-down.feature
```diff
@@ -0,0 +1,24 @@
+Feature: Marriage Broken Down
```
SebFlippence, line 1: This should be agnostic e.g.
```suggestion
Feature: Partnership Broken Down
```

linusnorton: maybe this should be a separate change as it doesn't relate to this PR? Also, the filename will need to be updated

src/test/functional/features/marriage-broken-down.feature
```diff
@@ -0,0 +1,24 @@
...
+
+  Background:
+    Given I am on divorce homepage
+    And I am on the admin portal sign in page
+    When I fill in the Username and Password fields with a valid login
+    And I click "Sign in"
+    And I see the divorce homepage
+    And I go to '/screening-questions/has-relationship-broken'
```
SebFlippence, line 9: This URL has been updated to match the requirements in #194
```suggestion
    And I go to '/irretrievable-breakdown'
```

velimd: will be fixed once i merge master

src/main/steps/index.test.ts
```diff
@@ -31,7 +36,7 @@ describe('Steps', () => {
 
     it('returns current step if there is no next step', () => {
       mockReq.originalUrl = HAS_RELATIONSHIP_BROKEN_URL;
-      expect(getNextStepUrl(mockReq)).toBe(HAS_RELATIONSHIP_BROKEN_URL);
+      expect(getNextStepUrl(mockReq)).toBe(RELATIONSHIP_DATE_URL);
```
SebFlippence, line 39: I think this test should be something like:

```ts
      const steps = getSteps();
      const lastUrl = steps[steps.length + 1];
      mockReq.originalUrl = lastUrl;
      expect(getNextStepUrl(mockReq)).toBe(lastUrl);
```

SebFlippence (approved):

PR #195, opened 2021-02-16
Save to API on form submit

linusnorton:
### JIRA link ###

https://tools.hmcts.net/jira/browse/NFDIV-460

### Change description ###

Improve API error handling, update the case using a PATCH request after submit, add Case definition to API responses


velimd (approved):

src/main/steps/sequence/irretrievable-breakdown/content.ts
```diff
@@ -2,9 +2,9 @@ import { TranslationFn } from '../../../app/controller/GetController';
 import { FormContent } from '../../../app/form/Form';
 import { isFieldFilledIn } from '../../../app/form/validation';
 
-export const generateContent = (title: string): TranslationFn => ({ isDivorce }) => {
+export const generateContent: TranslationFn = ({ isDivorce }) => {
   const en = {
-    title: isDivorce ? title : 'Has your relationship irretrievably broken down (it cannot be saved)?',
+    title: 'Has your relationship irretrievably broken down (it cannot be saved)?',
```
SebFlippence, line 7: You'll need a `isDivorce` title here

src/main/steps/urls.ts
```diff
@@ -12,7 +12,7 @@ export const ACCESSIBILITY_STATEMENT_URL: PageLink = '/accessibility-statement';
 
 export const YOUR_DETAILS_URL: PageLink = '/your-details';
 export const HAS_RELATIONSHIP_BROKEN_URL: PageLink = '/irretrievable-breakdown';
-export const RELATIONSHIP_NOT_BROKEN_URL: PageLink = '/you-cannot-apply';
+export const RELATIONSHIP_NOT_BROKEN_URL: PageLink = '/relationship-not-broken';
```
SebFlippence, line 15: Looking to the ticket requirements again for [NFDIV-441](https://tools.hmcts.net/jira/browse/NFDIV-441), this URL should be `/you-cannot-apply`

SebFlippence (changes requested):
Looks good other than a few minor comments (above)

src/main/steps/urls.ts
linusnorton: I'm going to push back on that because the page is specifically about the relationship and that page name is too generic

src/main/steps/sequence/irretrievable-breakdown/content.ts
linusnorton: I believe it was the same

linusnorton: It was not the same. I cleaned up the logic in there because it was pretty repetitive

SebFlippence (approved):

PR #228, opened 2021-02-23
Remove empty date check

linusnorton:
Remove unnecessary date check

### DIFF
--- a/src/main/app/api/to-api-format.ts
+++ b/src/main/app/api/to-api-format.ts
@@ -20,9 +20,6 @@ const fields = {
 };
 
 const toApiDate = (date: CaseDate) => {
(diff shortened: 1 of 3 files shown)


nitinprabhu (approved):

src/main/app/api/to-api-format.ts
```diff
@@ -20,9 +20,6 @@ const fields = {
 };
 
 const toApiDate = (date: CaseDate) => {
-  if (!date?.year || !date?.month || !date?.day) {
```
SebFlippence, line 23: Removing this might cause one of the functional tests to fail, as it will try to submit or save and sign out with a partially complete day/month/year field.

Perhaps we need to complete the "API failed to save" story to handle this situation.

PR #235, opened 2021-02-23
Remove awilix

linusnorton:
Remove awilix


src/main/modules/oidc/index.ts
```diff
@@ -60,33 +59,30 @@ export class OidcMiddleware {
...
-                baseURL: config.get('services.cos.baseURL'),
-                headers: {
-                  Authorization: 'Bearer ' + user.access_token,
-                  IdToken: user.id_token,
-                },
-              })
-            ),
-            api: asClass(CaseApi),
-          });
+          req.locals.api = new CaseApi(
```
velimd, line 65: im starting to think, is it best to create an api controller every time the user sends a request?

its not a big deal its just made me think if we should check before create maybe?

velimd (approved):

PR #238, opened 2021-02-23
Nfdiv 371 relationship not long enough

velimd:
### JIRA link ###

https://tools.hmcts.net/jira/browse/NFDIV-371

### Change description ###

```
[ ] Yes
[X] No
```


src/test/functional/features/relationship-date-formed.feature
```diff
@@ -30,6 +30,16 @@ Feature: Relationship date formed
     When I click "Continue"
     Then the page should include "You have entered a date that is in the future. Enter a date that is in the past before continuing."
 
+  Scenario: Entering date less than a year
+    Given I select "Day"
+    And I type "1"
+    Given I select "Month"
+    And I type "1"
+    Given I select "Year"
+    And I type "2021"
```
SebFlippence, line 39: Perhaps it might be better to make a new step that handles entering a dynamic date (e.g. 6 months ago). Rather than hard coding a date, since this will cause the test to failing next year.

velimd: yh i was thinking that but then i thought, is this going to repeatly going to break

SebFlippence (approved):

src/main/steps/index.ts
```diff
@@ -47,10 +47,10 @@ export const getNextIncompleteStepUrl = (req: AppRequest): string => {
   return `${url}${queryString}`;
 };
 
-export const getNextStepUrl = (req: AppRequest): string => {
+export const getNextStepUrl = (req: AppRequest, data: Partial<Case>): string => {
   const { path, queryString } = getPathAndQueryString(req);
   const nextStep = sequence.find(s => s.url === path);
-  const url = nextStep ? nextStep.getNextStep(req.body) : CHECK_ANSWERS_URL;
+  const url = nextStep ? nextStep.getNextStep(data) : CHECK_ANSWERS_URL;
```
SebFlippence, line 53: Looking at the tests, I don't think there's a need to change from `req.body` and add this unnecessary `data` param as it should always exist on `req` right?

velimd: i guess but it depends on the parsedbody return

SebFlippence (approved):

PR #240, opened 2021-02-23
Add service auth provider client

linusnorton:


src/main/app/auth/service/get-auth-token.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const initAuthToken = (): void => {
+  getTokenFromApi();
+  setInterval(getTokenFromApi, 1000 * 60 * 60);
+};
+
+export const getAuthToken = (): string => {
+  return token;
+};
```
velimd, line 30: is an auth token for the service and not the user?

velimd (approved):

src/main/app/auth/service/get-auth-token.ts
linusnorton: there are two, I should rename this method to be clearer: getServiceAuthToken

PR #241, opened 2021-02-23
fix: checkbox deselection

SebFlippence:
### Change description ###

Currently ticking a checkbox and continuing and then going back and deselecting a checkbox fails to update the `userCase`.

This PR will set a default value for checkboxes as an empty string (via a hidden field) and use a parser for checkboxes to return the correct result.

This also changes field parsers to be non-mutating.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/form/Form.ts
+++ b/src/main/app/form/Form.ts
@@ -7,13 +7,22 @@ export class Form {
   /**
    * Pass the form body to any fields with a parser and return mutated body;
    */
-  public getParsedBody(body: AnyObject): void {
-    Object.keys(this.form.fields)
-      .filter(key => this.form.fields[key].parser !== undefined)
-      .forEach(propertyName => {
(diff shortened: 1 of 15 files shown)


velimd (approved):

src/main/app/form/Form.ts
```diff
@@ -7,19 +7,28 @@ export class Form {
...
    */
-  public getParsedBody(body: AnyObject): void {
-    Object.keys(this.form.fields)
-      .filter(key => this.form.fields[key].parser !== undefined)
-      .forEach(propertyName => {
-        const field = <FormField & { parser: Parser }>this.form.fields[propertyName];
-        field.parser(body);
-      });
+  public getParsedBody(body: AnyObject): Partial<CaseWithFormData> {
+    const parsedBody = Object.entries(this.form.fields)
```
linusnorton, line 11: There's got to be a way to tidy this up

PR #247, opened 2021-02-25
Nfdiv 478 integrate ccd

linusnorton:
### JIRA link ###

https://tools.hmcts.net/jira/browse/NFDIV-478

### Change description ###

Switch COS API for CCD

### DIFF
--- a/src/main/app/case/CaseApi.test.ts
+++ b/src/main/app/case/CaseApi.test.ts
@@ -1,34 +1,35 @@
 import axios from 'axios';
 
-import { CaseApi } from './CaseApi';
+import { CaseApi, getCaseApi } from './CaseApi';
 import { CaseType } from './case';
 
-describe('CaseApi', () => {
-  const results = {
-    data: {
-      id: '1234',
-      error: null,
-      status: 'success',
-      allocatedCourt: {
-        court: 'court',
-      },
-      data: {
-        divorceOrDissolution: 'divorce',
-      },
-    },
-  };
+jest.mock('axios');
 
-  jest.mock('axios');
+const userDetails = {
+  token: '123',
+} as never;
+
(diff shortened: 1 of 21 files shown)


src/main/app/case/CaseApi.test.ts
```diff
@@ -1,52 +1,59 @@
...
     error: async (message: string) => message,
     info: async (message: string) => message,
   } as never;
-  const api = new CaseApi(mockedAxios, mockLogger);
+
+  const api = new CaseApi(mockedAxios, userDetails, mockLogger);
 
   test('Should return case data response', async () => {
-    mockedAxios.get = jest.fn().mockResolvedValue(results);
+    mockedAxios.get = jest.fn().mockResolvedValue({
```
SebFlippence, line 23: This reassignment of `axios.get` could have unintended side effect for other tests and Jest's automocking setup and teardown steps, as this looks like this is modifying axios globally. 

Jest's recommended way to change the mocked resolved value is to call methods directly on a mocked import:

```suggestion
    mockedAxios.get.mockResolvedValue({
```

src/main/app/case/CaseApi.test.ts
```diff
@@ -1,52 +1,59 @@
...
 
     expect(userCase).toStrictEqual({ id: '1234', divorceOrDissolution: 'divorce' });
   });
 
   test('Should return false when case was not found', async () => {
-    mockedAxios.get = jest.fn().mockRejectedValue({
-      response: {
-        status: 404,
-      },
+    mockedAxios.get = jest.fn().mockResolvedValue({
```
SebFlippence, line 40: Same as above looks like there are few of these can be replaced

SebFlippence (approved):

PR #276, opened 2021-03-02
Session Timeout Task

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-482
https://tools.hmcts.net/jira/browse/NFDIV-359

### Change description ###
- call '/active' endpoint when user is active to add to extend session maxage
- if user is inactive for 20mins, call post to save data and then redirect to timed-out endpoint

```
[ ] Yes
[X] No
```


src/main/app/controller/GetController.test.ts
```diff
@@ -19,6 +19,7 @@ describe('GetController', () => {
       extraEnglish: 'text',
       formState: req.session.userCase,
       hideBackButton: false,
+      hideNavigationButton: false,
```
linusnorton, line 22: shouldn't this default to false?

src/main/assets/js/session-timeout.ts
```diff
@@ -0,0 +1,41 @@
...
+  const url = window.location.pathname;
+  const body = Object.fromEntries(formData);
+  fetch(url, {
+    method: 'POST',
+    headers: {
+      'Content-type': 'application/json',
+      'csrf-token': body._csrf as string,
+    },
+    body: JSON.stringify(body),
+  }).then(response => {
```
linusnorton, line 20: maybe it would be better to make this async and use await?

src/main/app/controller/GetController.test.ts
velimd: default should be true as you want to see the sign in/out button but in this specific page, its not there hence defaut is true.

if youre talking about in the actual get class, it does compare to see if the url is '/timed-out'

src/main/assets/js/session-timeout.ts
velimd: just for my understanding, why would that be better? i dont have a preference really but just thought the .then block makes it easier to read and as its combined together.

linusnorton: No technical difference, just personal preference I guess. I prefer to stick to the `const someVariable = something()` look.

For example:

```
  const response = await fetch(url, {
    method: 'POST',
    headers: {
      'Content-type': 'application/json',
      'csrf-token': body._csrf as string,
    },
    body: JSON.stringify(body),
  });

  if (response.redirected) {
    window.location.href = response.url;
  }
```

linusnorton (approved):

src/main/steps/common/page.njk
```diff
@@ -12,12 +12,12 @@
...
     serviceName: pageHeader[serviceType],
     serviceUrl: (globals.basePath if globals.basePath) + "/",
-    navigation: [
+    navigation: ([
       {
         href: "/logout" if isLoggedIn else "/login",
         text: signOut if isLoggedIn else signIn
       }
-    ]
+    ]) if not hideNavigationButton
```
SebFlippence, line 20: Rather than scattering `hideNavigationButton` throughout the code, it might be better to leave this as it is and change the block in the affected template.njk so that these links don't get rendered. For example: [#281/src/main/steps/save-sign-out/template.njk](https://github.com/hmcts/nfdiv-frontend/pull/281/files#diff-2ccf0cd072b3a6562cd536fa373ceed4eefc8dafb09ae346ceba79024d6e0786R3-R13)

SebFlippence (approved):

src/main/assets/js/session-timeout.ts
```diff
@@ -0,0 +1,42 @@
+import _ from 'lodash';
```
SebFlippence, line 1: For webpack tree shaking you might want to change this to:

```suggestion
import { throttle } from 'lodash';
```

This way webpack should only bundle `throttle` to the frontend rather than including the whole of lodash.

src/main/app/controller/PostController.ts
```diff
@@ -16,20 +16,23 @@ export class PostController<T extends AnyObject> {
...
-    if (!isSaveAndSignOut && errors.length > 0) {
+    if (!isSaveAndSignOut && !isSessionTimeout && errors.length > 0) {
       req.session.errors = errors;
       nextUrl = req.url;
     } else {
       await req.locals.api.updateCase(req.session.userCase?.id, formData);
       if (isSaveAndSignOut) {
         return;
+      } else if (isSessionTimeout) {
+        return res.end();
```
SebFlippence, line 35: Perhaps this branch could be tested in `PostController.test`?

src/main/assets/js/session-timeout.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const pingUserActive = throttle(() => {
+  fetch('/active').then(() => {
+    setSaveTimeout();
+  });
+}, eventThrottleTimer);
+
+['click', 'touchstart', 'mousemove', 'keypress'].forEach(evt => document.addEventListener(evt, pingUserActive));
+setSaveTimeout();
```
SebFlippence, line 42: Would be nice to get some unit tests on this. I think you should be able to trigger a `keypress`, `click`, `mousemove` etc using Jest and then using Jest mock timers to advance the clock by 20 minutes you should be able to check if `fetch` was called with the right arguments.

velimd: this is becoming a bit of a pain hence i might merge this once you approve and do this in another pr

SebFlippence (approved):

PR #279, opened 2021-03-02
fix: back button + reset tests changing session language

SebFlippence:
### Change description ###

Fixes for:

* Back button hiding on the your details page via Nunjucks
* Functional test that change the session language, but don't change it back again
* Tests with duplicated names warnings
* Fix step sequence for help with fees steps

```
[ ] Yes
[x] No
```


velimd (commented):
how comes youre removing the backlink condition?

SebFlippence:
The logic still works in the same way, where the Back link is hidden on the first question and on the save and sign out pages.

But instead of passing around `hideBackButton` (e.g. pages that don't use it) it's now directly used on the pages that do need to customise/remove the Back link via a Nunjucks block: `{% block backLink %}Change/remove bank link{% endblock %}`

PR #308, opened 2021-03-08
fix: save & sign out get controller

SebFlippence:
### Change description ###

Some exit pages/dead ends where using a "Save and sign out" link which was moved to a post controller.

This PR fixes a 404 error with the "Save and sign out" link/request by adding the missing get controller.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/controller/GetController.ts
+++ b/src/main/app/controller/GetController.ts
@@ -24,7 +24,11 @@ export type TranslationFn = ({
 
 @autobind
 export class GetController {
-  constructor(protected readonly view: string, protected readonly content: TranslationFn | Translations) {}
+  constructor(
+    protected readonly view: string,
+    protected readonly content: TranslationFn | Translations,
+    protected language = 'en'
+  ) {}
 
   public async get(req: AppRequest, res: Response): Promise<void> {
     if (res.locals.isError || res.headersSent) {
--- a/src/main/app/controller/GetController.ts
+++ b/src/main/app/controller/GetController.ts
@@ -24,11 +24,7 @@ export type TranslationFn = ({
 
 @autobind
 export class GetController {
-  constructor(
-    protected readonly view: string,
-    protected readonly content: TranslationFn | Translations,
-    protected language = 'en'
-  ) {}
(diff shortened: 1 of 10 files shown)


src/main/app/controller/GetController.ts
```diff
@@ -22,7 +22,11 @@ export type TranslationFn = ({
 
 @autobind
 export class GetController {
-  constructor(protected readonly view: string, protected readonly content: TranslationFn | Translations) {}
+  constructor(
+    protected readonly view: string,
+    protected readonly content: TranslationFn | Translations,
+    protected language = 'en'
```
linusnorton, line 28: I don't see how we can set this at boot time.

SebFlippence: 👍 Good spot, I've updated this in 6b56ffd to make it clearer than `'en'` is being used as the default and then this gets updated there's a language set in the session. The session for these sign out pages will be destroyed, so this allows the child to update the parents language before it gets destroyed.

SebFlippence: 👍 Good spot, I've changed this so that pages with a "Save and sign out" link use the `?lng=` query string to keep their preference on these pages that clear out of the users session.

src/main/steps/get-certified-translation/template.njk
```diff
@@ -6,5 +6,5 @@
   <p class="govuk-body"> {{ line2 }}</p>
   <h3 class="govuk-heading-m">{{ subHeading }}</h3>
   <p class="govuk-body"> {{ line3 | safe }}</p>
-  <p class="govuk-body"><a href="/save-and-sign-out" class="govuk-link">{{ saveAndSignOut }}</a></p>
+  <p class="govuk-body"><a href="/save-and-sign-out?lng={{language}}" class="govuk-link">{{ saveAndSignOut }}</a></p>
```
velimd, line 9: might be best to have a middleware that adds the language?

velimd (approved):

linusnorton (approved):

PR #323, opened 2021-03-09
feat(NFDIV-510): Added check if you can get a divorce page

SebFlippence:
### JIRA link ###

[NFDIV-510](https://tools.hmcts.net/jira/browse/NFDIV-510) English version - Check if you can get a divorce / end a civil partnership in England or Wales

### Change description ###

Added check if you can get a divorce page.

```
[ ] Yes
[x] No
```


src/main/steps/check-jurisdiction/content.ts
```diff
@@ -0,0 +1,30 @@
...
+    line2: 'The following questions will find out what connections you have to England or Wales.',
+  };
+
+  // @TODO translations
+  const cy: typeof en = {
+    ...en,
+  };
+
+  const common = {
+    form,
```
linusnorton, line 19: Where is this defined? Does this page have any fields? If so they need to be mapped to the API

SebFlippence: `form` is defined just below, there are no fields for this page, but there is a "Continue" button. Having an empty `fields` object renders the page correctly without any fields and with the "Continue" button at the bottom.

linusnorton (approved):

PR #340, opened 2021-03-11
feat(NFDIV-493): Added Where are your lives based question

SebFlippence:
### JIRA link ###

[NFDIV-493](https://tools.hmcts.net/jira/browse/NFDIV-493) English version - Where your lives are based

### Change description ###

Added Where are your lives based question and show errors in field for pages that have more than one field.

```
[ ] Yes
[x] No
```


SarahLittlejohn (approved):

src/main/modules/nunjucks/index.ts
```diff
@@ -20,6 +20,22 @@ export class Nunjucks {
...
+      if (!hasMoreThanTwoFields || !sessionErrors.length) {
+        return false;
+      }
+
+      const fieldError = sessionErrors.find(error => error.propertyName === fieldName);
+      if (!fieldError) {
+        return false;
+      }
+
+      return { text: errors[fieldName][fieldError.errorType] };
```
velimd, line 36: should this not be sessionErrors?

not sure what errors is.

SebFlippence: On line 31 I'm checking to see if this field has an error using `sessionErrors`. So technically I am getting the error from `sessionErrors` for each field.

velimd: just confused on what errors object is as we only return sessionErrors in the getController

SebFlippence: `errors` is a language object is from the [`content.ts:12`](https://github.com/hmcts/nfdiv-frontend/pull/340/files#diff-945fc05018a3db2c23dab2c87c47229550f445c57676fdcd83cd8058471e8f6dR12), which contains the two error messages for the fields.

And according to TypeScript `sessionErrors` is:

```ts
type FormError = {
  propertyName: string;
  errorType: string;
};

type sessionErrors = FormError[] | undefined;
```

So I'm finding the field based on `propertyName === fieldName` and then using the `errorType` to get the corresponding field error message from it's content file.

velimd: ahh, forgot about errors object in content. all good

velimd (approved):

PR #346, opened 2021-03-15
README.md update for zsh terminal on older MacOS systems.

kasi-subramaniam:
README.md update for zsh terminal on older MacOS systems.


README.md
```diff
@@ -57,6 +57,9 @@ The application makes calls to
 [case orchestration service](https://github.com/hmcts/nfdiv-case-orchestration-service).
 Make sure you have this running to run the frontend e2e.
 
+Some older versions of MacOS would require zsh to be installed and setup. This could help to avoid seeing
```
nitinprabhu, line 60: I would say instead of this could you please update something like use zsh while executing scripts in bin folder.

SebFlippence, line 60: Maybe adding a link to a guide would be useful and I think this would apply to all OSs and not just macOS.

```suggestion
zsh is required as the [default shell](https://superuser.com/questions/362372/how-to-change-the-login-shell-on-mac-os-x-from-bash-to-zsh). This fixes
```

kasi-subramaniam: this has been addressed now.

kasi-subramaniam: zsh installation link added to the ReadMe

SebFlippence (approved):

PR #351, opened 2021-03-15
Nfdiv 300

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-300

### Change description ###

```
[ ] Yes
[X] No
```


src/main/steps/common/common.content.ts
```diff
@@ -150,6 +152,8 @@ const cy: typeof en = {
   },
   yes: 'Do',
   no: 'Naddo',
+  english: 'English',
+  welsh: 'Welsh',
```
SebFlippence, line 156: You might want to exclude these two until we have the correct translations, as there's a `...en` above that will be doing this.

We'll probably want to run it passed the Welsh translation unit, but looking at previous translation it should be something like:
```suggestion
  english: 'Saesneg',
  welsh: 'Cymraeg',
```

SebFlippence (approved):

PR #353, opened 2021-03-15
feat(NFDIV-530): null unreachable answers

SebFlippence:
### JIRA link ###

[NFDIV-530](https://tools.hmcts.net/jira/browse/NFDIV-530) Clear previously entered, but now unreachable answers

### Change description ###

Sets previously entered, but now unreachable answers to `null`

```
[ ] Yes
[x] No
```


velimd (approved):

src/main/app/controller/PostController.ts
```diff
@@ -63,6 +69,20 @@ export class PostController<T extends AnyObject> {
       res.redirect(nextUrl);
     });
   }
+
+  private getAnsweredUnreachableResponsesAsNull(req: AppRequest<T>): Partial<Case> {
```
linusnorton, line 73: stop shoving everything in the post controller :D

SarahLittlejohn (approved):

linusnorton (changes requested):
I think setting an answer should knock out other fields

velimd:
> I think setting an answer should knock out other fields

That would be a pain for a user who is almost done but they change something which might not effect any path such as date on certificate or date of birth, which will make them start all over again. At that point, i rather stay married.

linusnorton:
> > I think setting an answer should knock out other fields
> 
> At that point, i rather stay married.

All part of the plan

src/main/app/controller/PostController.ts
SebFlippence: 👍 Fixed in 07670ab.

linusnorton (approved):

PR #364, opened 2021-03-16
feat(NFDIV-475): Error handling while saving form

SebFlippence:
### JIRA link ###

[NFDIV-475](https://tools.hmcts.net/jira/browse/NFDIV-475) Better error handling with API errors

### Change description ###

Keeps the user on the same page with an error if there was an API error while saving valid data.

```
[ ] Yes
[x] No
```


src/main/app/case/to-api-format.test.ts
```diff
@@ -6,7 +6,7 @@ describe('to-api-format', () => {
   const results: Partial<Case> = {
     gender: Gender.MALE,
     sameSex: Checkbox.Checked,
-    relationshipDate: { year: '1900', month: '1', day: '4' },
+    relationshipDate: { year: '1900', month: '01', day: '04' },
```
velimd, line 9: does this matter? the example of a date shows single digits.

src/main/app/controller/PostController.ts
```diff
@@ -63,6 +70,26 @@ export class PostController<T extends AnyObject> {
...
+  private async saveAndUpdateSession(
+    req: AppRequest<T>,
+    formData: Partial<Case>,
+    eventName: string
+  ): Promise<CaseWithId | undefined> {
+    try {
+      const caseData = await req.locals.api.triggerEvent(req.session.userCase.id, formData, eventName);
+
+      for (const [key, value] of Object.entries(formData)) {
+        if (!isEqual(caseData[key], value ?? '')) {
```
velimd, line 83: why empty string?

src/main/app/form/parser.ts
```diff
@@ -8,7 +8,8 @@ export const covertToDateObject: DateParser = (property, body) =>
   ['day', 'month', 'year'].reduce(
     (newDateObj: CaseDate, date: string) => {
       const propertyName = `${property}-${date}`;
-      newDateObj[date] = body[propertyName];
+      newDateObj[date] =
+        body[propertyName] && date !== 'year' ? (body[propertyName] as string).padStart(2, '0') : body[propertyName];
```
velimd, line 12: i am assuming this has something to do with ccd and how info is stored but if so, they could be done in the toapiformat

src/test/functional/features/check-your-answers.feature
```diff
@@ -41,7 +41,8 @@ Feature: Check Your Answers
     And the page should include "Payment"
     And the page should include "Do you need help paying the fee for your divorce?	I need help paying the fee"
     And the page should include "Have you already applied for help with your divorce fee?	Yes"
-    And the page should include "HWF-ABC-123"
+    # @TODO
```
velimd, line 44: how comes this is commented out? might be more helpful for someone who picks it up later to have a comment on it.

velimd: same to the other ones.

src/main/app/case/to-api-format.test.ts
SebFlippence: To be consistent with what's stored in the API and for validating the current session against the API response, I've changed the `session.userCase` to match what should be stored in the API.

src/main/app/controller/PostController.ts
SebFlippence: This is required for an unchecked checkbox as it will be `undefined` in the form data.

src/main/app/form/parser.ts
SebFlippence: I've had to move this from the to-api-format, to make sure that the `session.userCase` closely matches what will be stored in the API.

This way when the user submits the "when did you get married" form, we can validate the data returned by the API making sure that it matches what's in the session to confirm that it has been saved correctly by the API.

src/test/functional/features/check-your-answers.feature
SebFlippence: 👍 This has been fixed in another PR which has been recently merged, so I've reverted these changes in 32740b2.

src/main/app/case/to-api-format.test.ts
linusnorton: I think it was this way because the frontend provides it as 1900, 1, 4 which the API doesn't support so we wanted to make sure padding was applied.

src/main/app/case/to-api-format.ts
```diff
@@ -34,23 +34,19 @@ const fields = {
...
   }),
   englishOrWelsh: (data: Case) => ({
     LanguagePreferenceWelsh: data.englishOrWelsh === LanguagePreference.Welsh ? YesOrNo.Yes : YesOrNo.No,
   }),
 };
 
 const toApiDate = (date: CaseDate | undefined) => {
-  if (!date?.year || !date?.month || !date?.day) {
+  const dateParts = ['year', 'month', 'day'];
+  if (!date || !dateParts.every(r => Object.keys(date).includes(r))) {
```
linusnorton, line 46: imho it was more readable before but I'm not too fussed

src/main/app/controller/PostController.ts
linusnorton: I don't think this should be here, can't it go in the fromApiFormat function?

velimd (approved):

src/main/app/case/to-api-format.ts
SebFlippence: 👍 I've swapped this back in a6e5eca.

src/main/app/controller/PostController.ts
SebFlippence: 👍 Good idea, I've moved this up into the `triggerEvent` in a6e5eca.

src/main/app/case/CaseApi.ts
```diff
@@ -71,13 +72,26 @@ export class CaseApi {
...
-      await this.axios.post(`/cases/${caseId}/events`, { event, data, event_token: tokenResponse.data.token });
+      const response = await this.axios.post(`/cases/${caseId}/events`, {
+        event,
+        data,
+        event_token: tokenResponse.data.token,
+      });
+      const caseData = { id: response.data.id, ...fromApiFormat(response.data.data) };
+
+      for (const [key, value] of Object.entries(userData)) {
+        if (!isEqual(caseData[key], value ?? '')) {
```
linusnorton, line 89: can this be handled by the fromApiFormat function?

src/main/app/controller/PostController.test.ts
```diff
@@ -108,33 +145,35 @@ describe('PostController', () => {
   test('Should save the users data and redirect to the next page if the form is valid with parsed body', async () => {
     getNextStepUrlMock.mockReturnValue('/next-step-url');
     const errors = [] as never[];
-    const body = { day: '1', month: '1', year: '2000' };
+    const body = { day: '11', month: '12', year: '2000' };
     const parsedBody = {
-      date: { day: '1', month: '1', year: '2000' },
+      date: { day: '11', month: '12', year: '2000' },
```
linusnorton, line 150: this seems like a bodge to avoid the 01 vs 1 issue

src/main/app/controller/PostController.ts
```diff
@@ -63,6 +69,14 @@ export class PostController<T extends AnyObject> {
       res.redirect(nextUrl);
     });
   }
+
+  private async save(req: AppRequest<T>, formData: Partial<Case>, eventName: string): Promise<CaseWithId | false> {
+    try {
+      return await req.locals.api.triggerEvent(req.session.userCase.id, formData, eventName);
+    } catch {
+      return false;
```
linusnorton, line 77: Need logging

src/main/app/form/Form.test.ts
```diff
@@ -130,8 +130,8 @@ describe('Form', () => {
     expect(form.getParsedBody(body)).toStrictEqual({
       field: 'YES',
       dateField: {
-        day: '1',
-        month: '1',
+        day: '01',
```
linusnorton, line 133: Still don't think this is right but I might not understand what the form is producing

src/main/app/case/to-api-format.test.ts
SebFlippence: 👍 I've refactored this so that the `fromApiFormat` will drop any padded zeros in 559b32b.

src/main/app/case/CaseApi.ts
SebFlippence: 👍 `value` was incorrectly being set to `undefined` when a checkbox was unchecked, I've fixed this in 9228f0a.

src/main/app/controller/PostController.test.ts
SebFlippence: 👍 I've refactored this so that the `fromApiFormat` will drop any padded zeros in 559b32b.

src/main/app/controller/PostController.ts
SebFlippence: `triggerEvent` will [catch any exceptions, log them, and rethrow](https://github.com/hmcts/nfdiv-frontend/blob/master/src/main/app/case/CaseApi.ts#L50) so we shouldn't need to do any logging here.

src/main/app/form/Form.test.ts
SebFlippence: 👍 I've refactored this so that the `fromApiFormat` will drop any padded zeros in 559b32b.

PR #370, opened 2021-03-17
Nfdiv 422

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-422

### Change description ###

```
[ ] Yes
[X] No
```


src/main/app/form/validation.ts
```diff
@@ -88,3 +88,9 @@ export const isInvalidHelpWithFeesRef: Validator = value => {
     }
   }
 };
+
+export const isFieldLetters: Validator = value => {
+  if (!(value as string).match(/^[a-zA-Z][a-zA-Z\s]*$/)) {
```
SebFlippence, line 93: I don't think you need the first `[a-zA-Z]` in this regex, do you?

```suggestion
  if (!(value as string).match(/^[a-zA-Z\s]*$/)) {
```

velimd: you do to accept uppercase letters.
the reason why i didnt use this `/^[a-zA-Z\s]*$/` is because this allows users to enter space only and it accepts it as its not empty.

SebFlippence: Ah I see, to exclude only spaces and also only match lowercase, uppercase and then spaces would this work?

```suggestion
  if (!(value as string).match(/^[^\s][a-zA-Z\s]*$/)) {
```

velimd: this will allow a number or special character in the beginning

SebFlippence (approved):

PR #386, opened 2021-03-23
Nfdiv 515 get connection

SarahLittlejohn:
**JIRA link**
https://tools.hmcts.net/jira/browse/NFDIV-515

**Change description**

- getConnection method created to find the jurisdiction connection
- getConnection method called on the POST to the jurisdiction/interstitial page
- connections field added to form data and CCD

**Does this PR introduce a breaking change? (check one with "x")**

```
[ ] Yes
[ X ] No
```


src/main/app/controller/PostController.ts
```diff
@@ -52,11 +53,18 @@ export class PostController<T extends AnyObject> {
       req.session.errors = errors;
       nextUrl = req.url;
     } else {
+      nextUrl = getNextStepUrl(req, req.session.userCase);
+      if (nextUrl === JURISDICTION_INTERSTITIAL_URL) {
```
linusnorton, line 57: rather than doing this here, find the field from that page and then do this logic in the to-api-format

src/main/app/case/definition.ts
```diff
@@ -148,6 +148,17 @@ export const enum ConfidentialAddress {
...
 }
+
+export const enum Connection {
+  A = 'A',
+  B = 'B',
+  C = 'C',
+  D = 'D',
+  E = 'E',
+  F = 'F',
+  G = 'G',
```
SebFlippence, line 159: Perhaps it might be nice to have some sort of human readable connection description, even if we're only sending A-G to the API e.g.

```suggestion
  RESIDUAL_JURISDICTION = 'G',
```

src/main/app/controller/PostController.ts
```diff
@@ -19,7 +19,10 @@ export class PostController<T extends AnyObject> {
    */
   public async post(req: AppRequest<T>, res: Response): Promise<void> {
     const { saveAndSignOut, saveBeforeSessionTimeout, _csrf, ...formData } = this.form.getParsedBody(req.body);
+    await this.postCase(req, res, formData);
+  }
 
+  public async postCase(req: AppRequest<T>, res: Response, formData: Partial<Case>): Promise<void> {
```
linusnorton, line 25: maybe protected?

src/main/routes.ts
```diff
@@ -50,7 +51,11 @@ export class Routes {
       app.get(step.url, errorHandler(controller.get));
 
       if (step.form) {
-        app.post(step.url, errorHandler(new PostController(new Form(step.form)).post));
+        if (jurisdictionSteps.map(String).includes(step.url)) {
```
linusnorton, line 54: I'll tidy this up in another PR so the step scanner detects whether there is a custom post controller defined.

src/main/app/case/to-api-format.ts
```diff
@@ -39,6 +40,104 @@ const fields = {
   englishOrWelsh: (data: Case) => ({
     LanguagePreferenceWelsh: data.englishOrWelsh === LanguagePreference.Welsh ? YesOrNo.Yes : YesOrNo.No,
   }),
+  partnersLifeBasedInEnglandAndWales: (data: Case) => {
+    const connection = addConnection(data);
+    if (connection) {
+      return {
+        JurisdictionConnections: [addConnection(data)],
```
linusnorton, line 47: why call addConnection twice? can you use just connection instead?

src/main/app/case/to-api-format.ts
```diff
@@ -39,6 +40,104 @@ const fields = {
...
+        JurisdictionConnections: [],
+        JurisdictionRespondentResidence: data.partnersLifeBasedInEnglandAndWales,
+      };
+    }
+  },
+  yourDomicileInEnglandWales: (data: Case) => {
+    const connection = addConnection(data);
+    if (connection) {
+      return {
+        JurisdictionConnections: [addConnection(data)],
```
linusnorton, line 61: as above

src/main/app/case/to-api-format.ts
```diff
@@ -39,6 +40,104 @@ const fields = {
...
+        JurisdictionResidualEligible: data.jurisdictionResidualEligible,
+      };
+    } else {
+      return {
+        JurisdictionConnections: [],
+        JurisdictionResidualEligible: data.jurisdictionResidualEligible,
+      };
+    }
```
linusnorton, line 139: all these functions follow the same pattern, we can simplify them all to

```
jurisdictionResidualEligible: (data: Case) => ({
  JurisdictionConnections: getConnections(data),
  JurisdictionResidualEligible: data.jurisdictionResidualEligible,
})
```

Also, why is JurisdictionConnections an array when it's only ever empty or one item?

src/main/steps/jurisdiction/interstitial/post.ts
```diff
@@ -0,0 +1,20 @@
+import autobind from 'autobind-decorator';
+import { Response } from 'express';
+
+import { AppRequest } from '../../../app/controller/AppRequest';
+import { AnyObject, PostController } from '../../../app/controller/PostController';
+import { Form } from '../../../app/form/Form';
+
+@autobind
+export class JurisdictionPostController extends PostController<AnyObject> {
```
linusnorton, line 9: I'm not sure what this class does, the behaviour looks the same as the parent class

src/main/app/case/to-api-format.ts
SarahLittlejohn: JurisdictionConnections is an array because it's going be used on the checkbox page so needs to be able to hold more than one connection.

And I tried to clean the functions up like that but was getting an error for trying to submit a null value into an array

src/main/steps/jurisdiction/interstitial/post.ts
SarahLittlejohn: Instead of posting the formData, it assigns the formData to req.session.userCase and posts that instead

src/main/app/case/to-api-format.ts
SarahLittlejohn: My bad - have just tried cleaning up the functions and it works !

linusnorton (approved):

PR #401, opened 2021-03-26
feat(NFDIV-454): Enter your postal address

SebFlippence:
### JIRA links ###

* [NFDIV-454](https://tools.hmcts.net/jira/browse/NFDIV-454) English version - Enter your postal address - Post code
* [NFDIV-464](https://tools.hmcts.net/jira/browse/NFDIV-464) English version - Enter your postal address - Cannot enter a UK post code
* [NFDIV-463](https://tools.hmcts.net/jira/browse/NFDIV-463) English version - Enter your address - Cannot find the address in list

### Change description ###

Adds postcode to full UK address lookup flow, including option to manually enter an international address or manually enter a UK address.

```
[ ] Yes
[x] No
```


src/main/app/case/case.ts
```diff
@@ -71,6 +72,13 @@ export interface Case {
   livingInEnglandWalesSixMonths?: YesOrNo;
   jurisdictionResidualEligible?: YesOrNo;
   englishOrWelsh?: LanguagePreference;
+  yourAddress1?: string;
```
linusnorton, line 75: I don't think we should have optionals in here

linusnorton: also, you can't use string here

src/main/app/case/case.ts
```diff
@@ -71,6 +72,13 @@ export interface Case {
   livingInEnglandWalesSixMonths?: YesOrNo;
```
linusnorton, line 72: I know this isn't your code but you can't have a type here. This index is a mapping of one to one fields between the API and the frontend

src/main/app/case/from-api-format.ts
```diff
@@ -15,6 +15,28 @@ const fields = {
     englishOrWelsh:
       data.LanguagePreferenceWelsh === YesOrNo.Yes ? LanguagePreference.Welsh : LanguagePreference.English,
   }),
+  D8DerivedPetitionerHomeAddress: data => {
```
linusnorton, line 18: lets move this out to a function somewhere

src/main/app/case/from-api-format.ts
```diff
@@ -15,6 +15,28 @@ const fields = {
...
+    if (!data.D8DerivedPetitionerHomeAddress) {
```
linusnorton, line 19: when does this happen?

src/main/app/form/validation.ts
```diff
@@ -89,6 +89,19 @@ export const isInvalidHelpWithFeesRef: Validator = value => {
   }
 };
 
+export const isInvalidPostcode: Validator = value => {
+  const fieldNotFilledIn = isFieldFilledIn(value);
+  if (fieldNotFilledIn) {
+    return fieldNotFilledIn;
+  }
+
+  if (typeof value === 'string') {
```
linusnorton, line 98: this value will definitely be a string if it is filled in so you don't need this check

src/main/app/services/postcodeLookup.ts
```diff
@@ -0,0 +1,66 @@
...
+        street2: [DEPENDENT_THOROUGHFARE_NAME, DEPENDENT_LOCALITY, DOUBLE_DEPENDENT_LOCALITY]
+          .filter(Boolean)
+          .join(', '),
+        town: POST_TOWN,
+        county: LOCAL_CUSTODIAN_CODE_DESCRIPTION,
+        postcode: POSTCODE,
+      })
+    );
+  } catch (err) {
+    if (err?.response?.status === StatusCodes.UNAUTHORIZED) {
```
linusnorton, line 61: when is err not set?

src/main/app/services/postcodeLookup.ts
```diff
@@ -0,0 +1,66 @@
...
+      logger.error('Postcode lookup key is invalid');
+    }
```
linusnorton, line 63: what about an else statement? you'll probably want to log other errors too

src/main/steps/common/form/fields.njk
```diff
@@ -69,6 +72,22 @@
         items: formItems(field.values, formState[fieldName]),
         errorMessage: getError(fieldName)
       }) }}
+    {% case 'textarea' %}
```
linusnorton, line 75: I wonder if we should split out each case into it's own file

src/main/steps/enter-your-address/content.ts
```diff
@@ -0,0 +1,68 @@
...
+export const form: FormContent = {
+  fields: {
+    yourAddressPostcode: {
+      type: 'text',
+      classes: 'govuk-label govuk-input--width-10',
+      label: l => l.postcode,
+      labelSize: null,
+      hideError: true,
+      validator: (value, form) => {
+        if (form?.myAddressIsInternational === Checkbox.Checked) {
```
linusnorton, line 31: you've used the form variable name twice here

linusnorton: also when is it not set? Are you just slapping ? on everything? :D

linusnorton (commented):
I was hoping it would be a bit more re-usable. It looks like every time we have an address field we're going to need to copy four pages over. I was hoping it would be a single page with the javascript rendering each of the different views inside the page. Is that too much for our FE JS to handle?

src/main/steps/enter-your-address/content.ts
```diff
@@ -27,8 +27,8 @@ export const form: FormContent = {
       label: l => l.postcode,
       labelSize: null,
       hideError: true,
-      validator: (value, form) => {
-        if (form?.myAddressIsInternational === Checkbox.Checked) {
+      validator: (value, form = {}) => {
```
linusnorton, line 30: when does this method get called without the form argument?

src/main/app/case/from-api-format.ts
SebFlippence: 👍 Good idea, I've moved this out into a seperate file

src/main/app/case/from-api-format.ts
SebFlippence: 👍 Good spot, this shouldn't, I've removed it

src/main/app/form/validation.ts
SebFlippence: 👍 Good point, I've cast this to a string

src/main/app/services/postcodeLookup.ts
SebFlippence: 👍 Good spot, it's always set

src/main/app/services/postcodeLookup.ts
SebFlippence: 👍 Good idea, I've updated this to catch everything else that's not a non existent postcode

src/main/steps/common/form/fields.njk
SebFlippence: 👍 Good idea, I've moved all of these out into seperate files

src/main/steps/enter-your-address/content.ts
SebFlippence: 👍 Good spot, I've renamed this and updated the types and other calls so I don't need to optionally chain this

src/main/steps/enter-your-address/content.ts
SebFlippence: 👍 Good shout, I've updated the types on this and other calls to it so I don't need to default it

SebFlippence (commented):
> I was hoping it would be a bit more re-usable. It looks like every time we have an address field we're going to need to copy four pages over. I was hoping it would be a single page with the javascript rendering each of the different views inside the page. Is that too much for our FE JS to handle?

I've updated this PR so it's now one single page where the client will handle the postcode lookup and update the address fields. I've made this reusable so that if you provide a form with a UK address and international address fields (and render out the postcode lookup and select address fields) then the client side code will pick these up and will automatically populate them.

linusnorton (approved):

PR #417, opened 2021-03-30
rename all the things

linusnorton:
Rename all the fields, update case type ID, make the API formatters have better type information

### DIFF
--- a/charts/nfdiv-frontend/values.yaml
+++ b/charts/nfdiv-frontend/values.yaml
@@ -20,8 +20,8 @@ nodejs:
     IDAM_WEB_URL: 'https://idam-web-public.{{ .Values.global.environment }}.platform.hmcts.net/login'
     IDAM_API_URL: 'https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net/o/token'
     REDIS_HOST: 'nfdiv-frontend-session-storage-{{ .Values.global.environment }}.redis.cache.windows.net'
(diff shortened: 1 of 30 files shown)


charts/nfdiv-frontend/values.yaml
```diff
@@ -20,7 +20,7 @@ nodejs:
     IDAM_WEB_URL: 'https://idam-web-public.{{ .Values.global.environment }}.platform.hmcts.net/login'
     IDAM_API_URL: 'https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net/o/token'
     REDIS_HOST: 'nfdiv-frontend-session-storage-{{ .Values.global.environment }}.redis.cache.windows.net'
-    CCD_URL: 'http://ccd-data-store-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal'
+    CCD_URL: 'http://ccd-data-store-api-nfdiv-case-api-pr-113.service.core-compute-preview.internal'
```
nitinprabhu, line 23: why do we need to hardcode pr number ?

linusnorton: I'll revert this before merging, but it links the frontend to the backend / CCD change so that the tests connect use the correct instance. If they used AAT all the field names would be wrong

nitinprabhu (approved):

SebFlippence (approved):

PR #419, opened 2021-03-31
Allow custom controllers for steps

linusnorton:


src/main/routes.ts
```diff
@@ -43,19 +42,16 @@ export class Routes {
...
-      app.get(step.url, errorHandler(controller.get));
+      app.get(step.url, errorHandler(new getController(view, step.generateContent).get));
 
       if (step.form) {
-        if (jurisdictionSteps.map(String).includes(step.url)) {
-          app.post(step.url, errorHandler(new JurisdictionPostController(new Form(step.form)).post));
-        } else {
-          app.post(step.url, errorHandler(new PostController(new Form(step.form)).post));
-        }
+        const postController = fs.existsSync(`${dir}/post.ts`) ? require(`${dir}/post.ts`).default() : PostController;
```
SarahLittlejohn, line 53: Does this work by looking into the directory and seeing if there is a custom post? I'm just thinking that there's 5 pages the custom post controller needs to be called on, but it lives inside the jurisdiction/interstitial directory

PR #423, opened 2021-03-31
Nfdiv 432

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-432

### Change description ###

```
[ ] Yes
[X] No
```


src/main/app/form/validation.ts
```diff
@@ -1,6 +1,8 @@
-import { CaseDate } from '../case/case';
+import * as EmailValidator from 'email-validator';
```
SebFlippence, line 1: Maybe this could be:
```suggestion
import { validate as validateEmail } from 'email-validator';
```

src/main/app/form/validation.ts
```diff
@@ -106,8 +108,29 @@ export const isPhoneNoValid: Validator = value => {
   }
 };
 
+export const isEmailValid: Validator = value => {
+  if (!EmailValidator.validate(value as string)) {
```
SebFlippence, line 112: Then, this could be:
```suggestion
  if (!isValidEmail(value as string)) {
```

SebFlippence (approved):

SebFlippence (approved):

PR #434, opened 2021-04-07
chore: handle multiple errors being caught

SebFlippence:
### JIRA link ###

N/A

### Change description ###

In some cases multiple errors are thrown and caught, in these cases the application was crashing due to headers already sent.

This PR checks if HTTP headers have already been sent before trying to render.

```
[ ] Yes
[x] No
```


linusnorton (approved):
Trust in Seb.

PR #435, opened 2021-04-07
feat(NFDIV-453): Added "You need to get their address page"

SebFlippence:
### JIRA link ###

[NFDIV-453](https://tools.hmcts.net/jira/browse/NFDIV-453) You need to get their address

### Change description ###

Added "You need to get their address page" and functional tests.

This also introduces an "is form complete" check where fields set as `undefined` or `null` are considered as unanswered/unseen and anything else including an empty string `""` (e.g. if it's optional) is considered as answered/visited.

Clicking "Save and sign out" on a step that has an unchecked checkbox will be considered as unanswered.

Steps in this PR can now be marked with an `excludeFromContinueApplication` which will use the previous step if this was the last step the user should be on before stopping/saving and signing out.

```
[ ] Yes
[x] No
```


src/main/app/case/to-api-format.ts
```diff
@@ -4,12 +4,22 @@ import { Case, CaseDate, Checkbox, LanguagePreference, formFieldsToCaseMapping,
...
 import { theirAddressToApi, yourAddressToApi } from './formatter/address';
 
-type ToApiConverters = Partial<Record<keyof Case, string | ((data: Case) => Partial<CaseData>)>>;
+export type OrNull<T> = { [K in keyof T]: T[K] | null };
+
+type ToApiConverters = Partial<Record<keyof Case, string | ((data: Case) => OrNull<Partial<CaseData>>)>>;
+
+const checkboxConverter = (value: string | undefined) => {
+  if (!value) {
+    return null;
```
velimd, line 13: would undefined not work?

velimd: that way we might not need the or null stuff. tho i guess its best.

src/main/app/controller/PostController.ts
```diff
@@ -31,6 +31,11 @@ export class PostController<T extends AnyObject> {
 
   private async saveAndSignOut(req: AppRequest<T>, res: Response, formData: Partial<Case>): Promise<void> {
     try {
+      for (const prop in formData) {
+        if (formData[prop] === Checkbox.Unchecked) {
+          formData[prop] = null;
```
velimd, line 36: i assume it has to be null and '' breaks CCD?

src/main/modules/nunjucks/index.ts
```diff
@@ -44,7 +44,7 @@ export class Nunjucks {
         classes: i.classes,
         value: i.value ?? userAnswer?.[i.name as string] ?? (userAnswer as string),
         attributes: i.attributes,
-        checked: i.selected ?? userAnswer?.[i.name as string] ?? i.value === userAnswer,
+        checked: i.selected || i.value === userAnswer?.[i.name as string] || i.value === userAnswer,
```
velimd, line 47: in multiple checkboxes situation, i use this userAnswer?.[i.name as string]?.includes(i.value as string) as the answers will be an array. will this work?

src/main/steps/common/form/fields/checkboxes.njk
```diff
@@ -2,23 +2,23 @@
 
 {% for checkbox in field.values %}
   {# Required for sending an empty value when the checkbox has not been checked #}
-  <input type="hidden" value="" name="{{ checkbox.name }}">
+  <input type="hidden" value="unchecked" name="{{ checkbox.name }}">
```
velimd, line 5: was there an issue with it being empty? just curious as in my multiple checkbox implementaiton, i get rid of empty values as this input value also is returned making it an array ['unchecked', 'checked'].

src/main/app/case/to-api-format.ts
SebFlippence: I've updated this in 4860c11 to explicitly check for `null`. The only time this will be null is when the checkbox hasn't been ticked and the user clicks "Save and sign out".

src/main/app/controller/PostController.ts
SebFlippence: I've moved this to the checkbox parser to only run on save and sign out. `null` means user has not answered this field, while `""` means they've not selected it.

src/main/modules/nunjucks/index.ts
SebFlippence: 👍 I've updated this to match #446

src/main/steps/common/form/fields/checkboxes.njk
SebFlippence: 👍 I've reverted this and kept `""` as unchecked.

velimd (approved):

PR #436, opened 2021-04-07
Nfdiv 249 do you have address

SarahLittlejohn:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-249

### Change description ###
Page implementation and tests for /do-you-have-address

**This branch is based off of the NFDIV-432 branch as it needs the answer on this page to drive the dynamic content.**

```
[ ] Yes
[ X ] No
```


src/main/app/form/validation.ts
```diff
@@ -106,8 +108,29 @@ export const isPhoneNoValid: Validator = value => {
...
+export const doesNotKnowEmail = (value: Partial<Case>): string | void => {
+  if (value.respondentEmailAddress && value.doNotKnowRespondentEmailAddress === Checkbox.Checked) {
+    return 'incorrect';
+  }
+};
+
+export const isEitherFieldsFilledIn = (value: Partial<Case>): string | void => {
+  if (
+    isFieldFilledIn(value.respondentEmailAddress) === 'required' &&
+    value.doNotKnowRespondentEmailAddress !== Checkbox.Checked
```
SebFlippence, line 132: Perhaps this method should be renamed to something like `isRespondentEmailFilledOrNotKnown` as `isEitherFieldsFilledIn` doesn't match what this is doing and sounds like it would take an array of fields to check.

src/main/steps/do-you-have-address/content.ts
```diff
@@ -0,0 +1,67 @@
...
+      : `It’s important you provide your ${partner}'s address. You did not provide their email address and the court needs a way to ‘serve’ (deliver) the ${
+          isDivorce ? 'divorce' : 'ending your civil partnership'
+        } papers to them.`
+  }`,
+  line2:
+    'It can be their home address or their solicitor’s address. It can be UK or international. If you use their work address, you need to ask their permission.',
+  line3: `${
+    partnerEmailProvided
+      ? 'If you do not know their current address then you can use their last-known address.'
+      : 'If you do not know their current address then you can use their last-known address. If they do not respond at the address you provide, then you will need to make a separate application to serve (deliver) the papers to them another way.'
```
SebFlippence, line 22: Looks like you could cut out a duplicated line here, by doing something like:

```suggestion
  line3: `If you do not know their current address then you can use their last-known address. ${
    !partnerEmailProvided
      ? 'If they do not respond at the address you provide, then you will need to make a separate application to serve (deliver) the papers to them another way.'
      : ''
```

src/main/app/form/validation.ts
SebFlippence: On second thoughts it might be better to move this out of the shared `validation.ts` file and into the `content.ts` it belongs to since it's not really sharable to any other question.

SarahLittlejohn: Ok cool, is it ok to still be tested inside validation.test or should I make a new file inside test/steps where we can test custom validation (assuming we'll probably have more in the future)?

SebFlippence: That's a good point. For the some of the more basic `content.ts` we've been relying on the functional tests to check that their working correctly, so if there's a functional test that covers this behaviour, then it may not be needed otherwise you can make `content.test.ts` next to it, to cover it.

The only downside with the functional tests is that the code coverage may go down even though technically there could be functional test that is validating this. The current best practices seem to suggest to [write more integration tests rather than unit tests](https://kentcdodds.com/blog/write-tests).

linusnorton: in this context functional tests would be e2e tests on his pyramid. We don't really have integration tests

src/main/steps/do-you-have-address/template.njk
```diff
@@ -0,0 +1,9 @@
+{% extends "common/page.njk" %}
+
+{% block page_content %}
+  <h1 class="govuk-heading-xl">{{ title }}</h1>
+  <p class="govuk-body">{{ line1 | safe }}</p>
+  <p class="govuk-body govuk-inset-text">
+      {{ line2 }}<br><br>{{ line3 }}
```
velimd, line 7: maybe make these p tags, not sure how the page looks tbh.

src/main/app/form/validation.ts
SarahLittlejohn: Looks like the functional tests cover this behaviour, so I'll see what the coverage looks like and if its too low I'll add in the unit test

SebFlippence (approved):

src/main/app/form/validation.ts
```diff
@@ -106,8 +108,20 @@ export const isPhoneNoValid: Validator = value => {
...
+};
+
 export const isFieldLetters: Validator = value => {
   if (!(value as string).match(/^[a-zA-Z][a-zA-Z\s]*$/)) {
     return 'invalid';
   }
 };
+
+export const doesNotKnowEmail = (value: Partial<Case>): string | void => {
+  if (value.respondentEmailAddress && value.doNotKnowRespondentEmailAddress === Checkbox.Checked) {
```
velimd, line 124: i feel like, if the other validation was moved due to being specific to the content, this should also be moved.

SarahLittlejohn: I guess you could say the same thing about  **isInvalidHelpWithFeesRef** too

PR #437, opened 2021-04-07
feat(NFDIV-536): Added "How to apply to serve" page

SebFlippence:
### JIRA link ###

[NFDIV-536](https://tools.hmcts.net/jira/browse/NFDIV-536) How to apply to serve

### Change description ###

This PR adds the "How to apply to serve" page and functional tests.

```
[ ] Yes
[x] No
```


src/main/app/case/case.ts
```diff
@@ -92,6 +92,7 @@ export interface Case {
   anyNameChangeSinceRelationshipFormed?: YesOrNo;
   changedNameHow?: ChangedNameHow;
   changedNameHowAnotherWay?: string;
+  respondentEmailAddress?: string;
```
velimd, line 95: how comes this is added here? is it something missed from another ticket as i dont see it being needed or used for this ticket.

velimd (approved):
just that one field added to the case object which seems out of place but other than that, all good.

src/main/app/case/case.ts
SebFlippence: Ah, this is needed for the the dynamic content:

```ts
formState?.respondentEmailAddress ? 'the email address you entered earlier' : 'email, text message or social media'
```

However, it since this field hasn't been implemented yet, we'll have to remember to double check that this field name is still correct once the ticket for their partners email has been merged.

PR #446, opened 2021-04-09
Nfdiv 302

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-302

### Change description ###
Note: This does require the field name to equal the name assigned to the object, as ive changed the parser to use that to find the formdata instead of looping around each value.
I believe the names should be the same for checkboxes as i see this a values which need to be group. if another check box was used, it should be another field imo but i guess we can discuss this and revert my changes to the parser to loop through the values.

```
[ ] Yes
[X] No
```


src/main/app/form/Form.test.ts
```diff
@@ -23,19 +23,15 @@ describe('Form', () => {
...
+      checkboxes: {
         type: 'checkboxes',
+        validator: isFieldFilledIn,
         values: [
-          { name: 'optionalCheckbox', label: () => 'optional', value: Checkbox.Checked },
-          {
-            name: 'requiredCheckbox',
-            label: () => 'required checkbox',
-            value: Checkbox.Checked,
-            validator: isFieldFilledIn,
```
SebFlippence, line 37: Could we keep a test for a required checkbox as there are a few places where the user is required to tick a checkbox.

SebFlippence: Unless checkbox individual checkbox validation is no longer a thing, and it doing it on the checkbox group?

SebFlippence (approved):

src/main/app/form/Form.test.ts
velimd: the method still works, i did move it outside the individual values but ill add it to the test as well.

velimd: [line 31](https://github.com/hmcts/nfdiv-frontend/pull/446/files#diff-f1e9f6cea5d7aff8749494ca8672015defeb8afeaefd59eebb16d7222fa1fcbbR31)

SarahLittlejohn (approved):

SebFlippence (approved):

src/test/functional/features/other-court-cases.feature
```diff
@@ -0,0 +1,31 @@
...
+    Given I clear the form
+    When I click "Continue"
+    Then the page should include "There was a problem"
+
+  Scenario: Successfully completing the form
+    Given I select "No"
+    When I click "Continue"
+    Then the page URL should be "/money-property"
+
+  Scenario: Successfully completing the form
```
SebFlippence, line 19: Reusing the same scenario name generates a warning in the console, perhaps we should use a different scenario name for these two.

SebFlippence (approved):

PR #453, opened 2021-04-12
fix: handlle missing session data when service-auth-token errors

SebFlippence:
### JIRA link ###

N/A

### Change description ###

Handles case were session is set, but fails to fetch user/userCase e.g. because of an service-auth-token error, or network error etc.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/steps/common/common.content.ts
+++ b/src/main/steps/common/common.content.ts
@@ -166,7 +166,7 @@ export const generatePageContent = ({
   pageContent?: TranslationFn;
   isDivorce?: boolean;
   formState?: Partial<Case>;
-  userEmail: string;
+  userEmail?: string;
 }): PageContent => {
   const commonTranslations: typeof en = language === 'en' ? en : cy;
   const serviceName = getServiceName(commonTranslations, isDivorce);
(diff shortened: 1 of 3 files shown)


velimd (commented):
not sure what has changed to achieve what is in the description.

SebFlippence:
> not sure what has changed to achieve what is in the description.

If you start the app without being connected to the VPN for example and try and sign in, you'll be signed in correctly by IDAM, but then get an error that we not able to connect the the auth service to verify and fetch the users case. In this scenario we show a 500 error page, but then another error occurs on the error page because we're unable to access `req.session?.user.email` as the user has a session but no user profile and case loaded, so the fix was also check they have a user profile via `req.session?.user?.email`.

This also adds a test to check that error page works even when the user profile and/or case is `undefined`.

velimd:
> > not sure what has changed to achieve what is in the description.
> 
> If you start the app without being connected to the VPN for example and try and sign in, you'll be signed in correctly by IDAM, but then get an error that we not able to connect the the auth service to verify and fetch the users case. In this scenario we show a 500 error page, but then another error occurs on the error page because we're unable to access `req.session?.user.email` as they have a partial user session but no user profile and case loaded, so the fix was also check they have a user profile via `req.session?.user?.email`.
> 
> This also adds a test to check that error page works even when the user profile and/or case is `undefined`.

i thought idam response had the email in the jwt token.

velimd (approved):

PR #458, opened 2021-04-13
Nfdiv 304 apply claim costs

SarahLittlejohn:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-304

### Change description ###
FT and page implementation for /apply-claim-costs

```
[ ] Yes
[ X ] No
```


src/main/app/case/case.ts
```diff
@@ -35,6 +35,7 @@ export const formFieldsToCaseMapping: Partial<Record<keyof Case, keyof CaseData>
   changedNameHowAnotherWay: 'petitionerNameChangedHowOtherDetails',
   respondentEmailAddress: 'respondentEmailAddress',
   knowPartnersAddress: 'petitionerKnowsRespondentsAddress',
+  claimCosts: 'divorceCostsClaim',
```
SarahLittlejohn, line 38: Not a fan of this variable being named 'divorceCostsClaim' in the api but it's being used in the solicitor journey - maybe we need a task to make the variable names apply for both div and dissolution of cp?

velimd: there is of cases like this, for example nfdiv 302 uses marriage as one of its options.

velimd (approved):

src/main/app/case/case.ts
SarahLittlejohn: Yeah I've seen a few, thats why I thought it might be better to change them all in a pr rather than just this field for this ticket

SebFlippence (approved):

PR #470, opened 2021-04-15
feat(NFDIV-557): Check your answers prayer & truth confirmation

SebFlippence:
### JIRA link ###

[NFDIV-557](https://tools.hmcts.net/jira/browse/NFDIV-557) Check your answers

### Change description ###

This PR adds:

* The prayer & truth confirmation checkboxes to the check your answers page
* Grouping of jurisdiction section into one "How you’re connected to England and Wales" question
* Reordering of questions on check your answers page
* Unit and functional tests

```
[ ] Yes
[x] No
```


src/main/app/case/answers/getAnswerRows.ts
```diff
@@ -27,17 +28,43 @@ export const getAnswerRows = function (section: Sections): GovUkNunjucksSummary[
...
+          ...generatePageContent({
+            language,
+            pageContent: step.generateContent,
+            isDivorce,
+            formState: processedFormState,
+            userEmail,
+          }),
+        };
+      } catch {
+        // Some steps may throw an error if the user is not allowed to view them yet
```
velimd, line 44: didnt know we had pages like this as so far, i just enter the url and it works, even in the functional test, we just enter the url.

src/main/steps/check-your-answers/form.njk
```diff
@@ -0,0 +1,9 @@
+{% extends "common/form/form.njk"%}
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+
+{% block after_fields %}
```
velimd, line 4: how comes this is hear and not in the template? i assume due to positioning?

src/main/steps/you-can-use-english-welsh-courts/content.ts
```diff
@@ -214,7 +214,13 @@ const cy = ({ isDivorce, partner }: CommonContent, connections: JurisdictionConn
 };
 
 export const form: FormContent = {
-  fields: {},
+  fields: {
+    connections: {
+      type: 'hidden',
+      label: l => l.title,
```
velimd, line 220: i assume this is only for the check your answers page to display a title?

velimd (approved):

src/main/steps/you-can-use-english-welsh-courts/content.ts
SebFlippence: Yeah, this input is be hidden on this page, but will be shown with a title on the check your answers page.

src/main/steps/check-your-answers/form.njk
SebFlippence: Nunjuck blocks do not inherit from one another, so you cannot use a block within a block and have to override them.

src/main/app/case/answers/getAnswerRows.ts
SebFlippence: The "You can use courts in England/Wales" page checks if they have a valid connection to England/Wales before being displayed otherwise the dynamic content will generate an empty page (if they've not completed the jurisdiction section).

src/main/steps/check-your-answers/form.njk
velimd: but what about the template page fro check your answers, can you not override it there?

SebFlippence: Blocks can only be customised from their matching extended file. As the block `after_fields` is only in `common/form/form.njk` it cannot be customised in the page `template.njk`, as blocks are not customisable/inherited from imported files so has to be created as a separate file to be able to customise them.

SarahLittlejohn (approved):

PR #483, opened 2021-04-19
Nfdiv 504

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-504

### Change description ###

```
[ ] Yes
[X] No
```


SebFlippence (changes requested):
This looks like a duplication of the `/you-can-use-english-welsh-courts` page. 

Is it possible add the content to the existing `you-can-use-english-welsh-courts/content.ts` using the `formState.connections` to choose the correct jurisdiction/connection message?

src/main/steps/sequence.ts
```diff
@@ -314,6 +323,10 @@ export const sequence: Step[] = [
     showInSection: Sections.Costs,
     getNextStep: () => UPLOAD_YOUR_DOCUMENTS,
   },
+  {
+    url: JURISDICTION_CONNECTION_SUMMARY,
+    getNextStep: () => CHECK_ANSWERS_URL,
```
SebFlippence, line 328: Shouldn't this go to the `CERTIFICATE_NAME` after this question?

src/main/steps/sequence.ts
```diff
@@ -195,8 +196,16 @@ export const sequence: Step[] = [
...
+      } else if (
+        data.yourLifeBasedInEnglandAndWales === YesOrNo.NO &&
+        data.partnersLifeBasedInEnglandAndWales === YesOrNo.NO &&
+        data.yourDomicileInEnglandWales === YesOrNo.NO &&
+        data.partnersDomicileInEnglandWales === YesOrNo.NO &&
+        data.lastHabituallyResident === YesOrNo.NO
+      ) {
         return JURISDICTION_MAY_NOT_BE_ABLE_TO;
+      } else if (data.lastHabituallyResident === YesOrNo.NO) {
+        return JURISDICTION_CONNECTION_SUMMARY;
```
SebFlippence, line 208: `lastHabituallyResident === YesOrNo.NO`, then going to `JURISDICTION_CONNECTION_SUMMARY` says: `both you and your ${partner} are habitually resident`.

I'm not sure if that matches what they have selected?

src/main/steps/connection-summary/content.ts
```diff
@@ -0,0 +1,39 @@
+import { TranslationFn } from '../../app/controller/GetController';
+
+const en = ({ partner }) => ({
+  title: 'You can use English or Welsh courts to get a divorce',
+  line1: 'Your answers indicate that you can apply for a divorce in England and Wales because:',
+  line2: `both you and your ${partner} are habitually resident`,
```
SebFlippence, line 6: Is it possible that this go into `you-can-use-english-welsh-courts/content.ts`, then you can reuse `enHabitualAndDomicileHelp`.

src/main/steps/sequence.ts
velimd: the next page is your name, but i believe that hasnt been implemented yet as we did the story refinement for it today

velimd:
> This looks like a duplication of the `/you-can-use-english-welsh-courts` page.
> 
> Is it possible add the content to the existing `you-can-use-english-welsh-courts/content.ts` using the `formState.connections` to choose the correct jurisdiction/connection message?

yhh, i should take this.

src/main/steps/connection-summary/content.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const en = ({ isDivorce, applyForDivorce, applyForDissolution, partner, formState }: CommonContent) => {
+  const apply = isDivorce ? applyForDivorce : applyForDissolution;
+  let reasonsToApplyInEnglandOrWales;
+  if (formState?.jurisdictionResidualEligible === YesOrNo.YES) {
+    reasonsToApplyInEnglandOrWales = 'the courts of England and Wales have jurisdiction on a residual basis';
+  } else if (formState?.lastHabituallyResident === YesOrNo.NO && formState?.sameSex === Checkbox.Checked) {
+    reasonsToApplyInEnglandOrWales = `your ${partner} is domiciled in England or Wales`;
+  } else {
+    reasonsToApplyInEnglandOrWales = 'you are domiciled in England and Wales';
```
SebFlippence, line 16: Could this logic be moved into `src/main/app/jurisdiction/connections.ts` or the existing logic be reused? So that we have one source of truth for how their connections to England/Wales are matched.

This way we can also then have an array of multiple connections for a multiple bullet list on the `/you-can-use-english-welsh-courts`

velimd: after decision, it seems the page is different even if some content is similar and later on will have radio components.

SebFlippence (approved):

PR #536, opened 2021-05-04
content update

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-568

### Change description ###

```
[ ] Yes
[X] No
```


SarahLittlejohn (approved):

src/main/steps/apply-for-help-with-fees/content.ts
```diff
@@ -10,9 +10,11 @@ const en = ({ isDivorce }) => ({
   line3: `After you have applied then you will receive a Help With Fees reference number. You should enter the reference number when you return to this ${
     isDivorce ? 'divorce' : 'ending a civil partnership'
   } application.`,
+  line4: 'If you have a Help With Fees reference number then you can <a href="/help-with-your-fee">enter it here.</a>',
```
SebFlippence, line 13: Perhaps it might be better to use the existing constant for this so that we don't need to update it in 2 places:

```suggestion
  line4: `If you have a Help With Fees reference number then you can <a href="${urls.HELP_WITH_YOUR_FEE_URL}">enter it here.</a>`,
```

SebFlippence (approved):

PR #542, opened 2021-05-05
chore(NFDIV-608): Handle AddressGlobalUK field types

SebFlippence:
### JIRA link ###

[NFDIV-608](https://tools.hmcts.net/jira/browse/NFDIV-608) Change Address fields to use built in types, and update field IDS

### Change description ###

Updates the data sent to and received from the API to match the `AddressGlobalUK` field type.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/case/formatter/address.ts
+++ b/src/main/app/case/formatter/address.ts
@@ -1,68 +1,75 @@
 import { Case } from '../case';
-import { CaseData, YesOrNo } from '../definition';
+import { AddressGlobalUK, CaseData, YesOrNo } from '../definition';
 
 export const fromApi = (data: Partial<CaseData>, address: 'your' | 'their'): Partial<Case> => {
   const isPetitionerAddress = address === 'your';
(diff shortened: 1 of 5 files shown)


src/main/app/case/formatter/address.ts
```diff
@@ -1,68 +1,75 @@
...
+    [`${address}InternationalAddress`]: isAddressInternational
+      ? [
+          fullAddress?.AddressLine1,
+          fullAddress?.AddressLine2,
+          fullAddress?.AddressLine3,
+          fullAddress?.PostTown,
+          fullAddress?.County,
+          fullAddress?.PostCode,
+          fullAddress?.Country,
+        ].join('\n')
```
linusnorton, line 21: why are international addresses combined like that?

linusnorton: I see, you reverse it before sending to the API

linusnorton (approved):

PR #558, opened 2021-05-07
fix: new user first page regression

SebFlippence:
### JIRA link ###

N/A

### Change description ###

This fixes a redirect issue to the "Check your answers" page for new users.

New users should see the first question when they first login.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/steps/home/get.ts
+++ b/src/main/steps/home/get.ts
@@ -1,8 +1,9 @@
 import { Response } from 'express';
 
-import { CaseApi } from '../../app/case/CaseApi';
 import { AppRequest } from '../../app/controller/AppRequest';
+import { Form } from '../../app/form/Form';
 import { CHECK_ANSWERS_URL, YOUR_DETAILS_URL } from '../../steps/urls';
+import { form as firstQuestionFormContent } from '../your-details/content';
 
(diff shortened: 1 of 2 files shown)


src/main/steps/home/get.ts
```diff
@@ -1,19 +1,21 @@
...
 export class HomeGetController {
   public get(req: AppRequest, res: Response): void {
     if (req.session.userCase.divorceOrDissolution !== res.locals.serviceType) {
       throw new Error('Invalid case type');
     }
 
-    const isCasePartiallyComplete = Object.entries(req.session.userCase).some(
-      ([key, value]) => !CaseApi.SPECIAL_FIELDS.includes(key) && value
-    );
+    const isFirstQuestionComplete = () => {
```
linusnorton, line 14: why is this a function?

velimd: i think its to do with how a new user will be redirected to the check your answers page as when they sign up, the name and lastname are already there so it doesnt go to the first question as case is partially complete.

SebFlippence: 👍 Good point, I've remove that as it's unnecessary

velimd (approved):

PR #595, opened 2021-05-17
chore: functional tests - set case directly from fixtures

SebFlippence:
### JIRA link ###

N/A

### Change description ###

Set the case data directly from fixtures, rather than jumping through all of the question pages to setup the case correct and also reduces the `waitForAction` to 500ms.

This reduces to the time taken to run from around 344 seconds to 132 seconds.

```
[ ] Yes
[x] No
```


linusnorton (approved):
Very nice

PR #606, opened 2021-05-19
NFDIV-703 Update with new marriage fields

linusnorton:


nitinprabhu (approved):

SebFlippence (commented):
Using the term "marriage" as a label/key is not valid for civil partnership applications. Anyone working on the application/looking at backend data could be confused, as it looks like it would be part of a divorce case rather than a dissolution.

linusnorton:
I agree but this is the approach our BA has gone with

PR #617, opened 2021-05-20
fix: production node-ts memory usage

SebFlippence:
### JIRA link ###

N/A

### Change description ###

Reduces production `ts-node` memory usage from around ~200 MiB to around ~70 MiB.

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   },
   "scripts": {
     "prepare": "husky install",
-    "start": "NODE_ENV=production ts-node -r tsconfig-paths/register src/main/server.ts",
+    "start": "NODE_ENV=production ts-node --transpile-only -r tsconfig-paths/register src/main/server.ts",
     "start:dev": "NODE_ENV=development yarn start:watch",
     "start:docker": "NODE_ENV=docker yarn start:watch",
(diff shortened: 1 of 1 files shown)


linusnorton:
Won't we lose the typescript errors though?

damongreen123:
> Won't we lose the typescript errors though?
Do you mean in the logs or the app @linusnorton ? the logs are empty.

SebFlippence:
> Won't we lose the typescript errors though?

`yarn start` is used on production. So by this point it would have gone through `yarn start:dev` used locally, which will still show TypeScript errors. There's also the Jenkins build that runs `yarn build:prod` which would fail the build if there are any TypeScript errors.

`--transpile-only` looks like it does ignore type errors, but [reduces RAM by using a faster transpile module](https://medium.com/aspecto/ts-node-ram-consumption-12c257e09e13).

linusnorton:
I was thinking about the stack trace, if we still have the stack trace then it's all good. Worth checking though!

linusnorton (approved):

SebFlippence:
> I was thinking about the stack trace, if we still have the stack trace then it's all good. Worth checking though!

Logs, errors and stack traces are still reported (just the TypeScript warning/errors will be stripped, but they get checked by CI anyway :tada:)

PR #624, opened 2021-05-21
update content

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-700
https://tools.hmcts.net/jira/browse/NFDIV-699

### Change description ###

```
[ ] Yes
[x] No
```


src/main/steps/sequence.ts
```diff
@@ -126,13 +126,10 @@ export const sequence: Step[] = [
...
-      if (data.applicationType === ApplicationType.JOINT_APPLICATION) {
-        return CHECK_ANSWERS_URL;
-      } else {
-        return data.applicant1HelpPayingNeeded === YesOrNo.YES ? HELP_PAYING_HAVE_YOU_APPLIED : IN_THE_UK;
-      }
-    },
+    getNextStep: data =>
+      data.applicationType === ApplicationType.JOINT_APPLICATION || data.applicant1HelpPayingNeeded === YesOrNo.YES
+        ? HELP_PAYING_HAVE_YOU_APPLIED
+        : IN_THE_UK,
```
lara-mpt, line 132: I think it could just be 
```suggestion
getNextStep: data => (data.applicant1HelpPayingNeeded === YesOrNo.YES ? HELP_PAYING_HAVE_YOU_APPLIED : IN_THE_UK),
```

lara-mpt: That's what's implied by my ticket - https://tools.hmcts.net/jira/browse/NFDIV-654

velimd: true, looking at both journeys on the miro board, both go to the same place so we dont need to check. i just continued from what was there.

src/main/app/controller/PostController.ts
```diff
@@ -62,7 +62,8 @@ export class PostController<T extends AnyObject> {
     if (req.session.errors.length === 0) {
       try {
         req.session.userCase = await this.save(req, formData, CITIZEN_UPDATE);
-      } catch {
+      } catch (err) {
+        req.locals.logger.error('Error when saving');
```
SebFlippence, line 66: Might be worth capturing the error as well:

```suggestion
        req.locals.logger.error('Error saving', err);
```

src/main/app/controller/PostController.test.ts
```diff
@@ -130,6 +133,7 @@ describe('PostController', () => {
 
     expect(getNextStepUrlMock).not.toHaveBeenCalled();
     expect(res.redirect).toBeCalledWith('/request');
+    expect(logger.error).toBeCalled();
```
SebFlippence, line 136: This could then be:

```suggestion
    expect(logger.error).toBeCalledWith('Error saving', 'Error saving');
```

SebFlippence (approved):

PR #645, opened 2021-05-26
extend cookie expiration before saving session manually.

velimd:
### Change description ###
There seems to be an issue with connect-redis and express-session where if you manually save the session after the original maxAge the session was created, the user data is destroyed, hence i updated the expires property before the save method is called.

Update: I believe the issue is, even though we set `rolling: true`, it seems like it does not update in the redis store when calling `/active`, therefore when submitting a post request, it fails due to checking the expiration of the session data in the store.

https://github.com/tj/connect-redis/issues/292

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/routes.ts
+++ b/src/main/routes.ts
@@ -8,6 +8,7 @@ import { DocumentManagerController } from './app/controller/DocumentManagementCo
 import { GetController } from './app/controller/GetController';
 import { PostController } from './app/controller/PostController';
 import { Form } from './app/form/Form';
+import { cookieMaxAge } from './modules/session';
 import { stepsWithContent } from './steps';
 import { AccessibilityStatementGetController } from './steps/accessibility-statement/get';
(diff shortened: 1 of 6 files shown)


src/main/routes.ts
```diff
@@ -80,7 +81,14 @@ export class Routes {
         if (!req.session.user) {
           return res.redirect(SIGN_OUT_URL);
         }
-        return res.end();
+        req.session.cookie.expires = new Date(Date.now() + cookieMaxAge);
+        req.session.cookie.maxAge = cookieMaxAge;
```
SebFlippence, line 85: Would it be better to have these 2 options set/moved into `src/main/modules/session/index.ts:26` or does that cause issues?

velimd: hmmm, maybe, the idea was to have it here as active was meant to extend the session when the user was active.

SebFlippence (approved):

PR #650, opened 2021-05-27
definition update

velimd:
### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/app/case/definition.ts
+++ b/src/main/app/case/definition.ts
@@ -1,6 +1,6 @@
 /* tslint:disable */
 /* eslint-disable */
-// Generated using typescript-generator version 2.31.861 on 2021-05-25 14:03:46.
+// Generated using typescript-generator version 2.31.861 on 2021-05-27 14:36:54.
 
 export interface Address {
   AddressLine1: string;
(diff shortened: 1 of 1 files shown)


lara-mpt (approved):

src/main/app/case/definition.ts
```diff
@@ -161,8 +161,8 @@ export interface CaseData {
   applicant1KnowsApplicant2Address: YesOrNo;
   applicant2HomeAddress: AddressGlobalUK;
   legalProceedings: YesOrNo;
-  legalProceedingsDetails: string;
-  legalProceedingsRelated: LegalProceedingsRelated[];
+  legalProceedingsByCase: ListValue<LegalProceeding>[];
+  legalProceedingsOther: string;
```
lara-mpt, line 165: Won't the implementations of these fields need their name changing too?

velimd: yep

velimd: its become more complex hence i closed this pr. Whats it requires is more than what we can give so i closed it as Tom (who made the changes) said he'll look into it.

PR #651, opened 2021-05-27
NFDIV-688: Migrate fields to frontend

tomxelliott:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/app/case/case.ts
+++ b/src/main/app/case/case.ts
@@ -10,7 +10,7 @@ import {
   FinancialOrderFor,
   Gender,
   JurisdictionConnections,
-  LegalProceedingsRelated,
+  LegalProceeding,
   ListValue,
   State,
   YesOrNo,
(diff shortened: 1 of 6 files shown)


src/main/app/case/case.ts
```diff
@@ -55,8 +55,8 @@ export const formFieldsToCaseMapping: Partial<Record<keyof Case, keyof CaseData>
   applicant2EmailAddress: 'applicant2EmailAddress',
   applicant1KnowsApplicant2Address: 'applicant1KnowsApplicant2Address',
   legalProceedings: 'legalProceedings',
-  legalProceedingsDetails: 'legalProceedingsDetails',
-  legalProceedingsRelated: 'legalProceedingsRelated',
+  legalProceedingsByCase: 'legalProceedingsByCase',
```
velimd, line 58: I believe you also need to update the names on the content page.

src/main/app/case/case.ts
```diff
@@ -140,8 +140,8 @@ export interface Case {
   applicant1KnowsApplicant2Address?: YesOrNo;
   iWantToHavePapersServedAnotherWay?: Checkbox;
   legalProceedings?: YesOrNo;
-  legalProceedingsDetails?: string;
-  legalProceedingsRelated?: LegalProceedingsRelated[];
+  legalProceedingsByCase?: ListValue<LegalProceeding>[];
```
velimd, line 143: this might be a problem, as LegalProceeding is an object which also has the properties CaseNumber/CaseDetails.

src/main/app/case/to-api-format.ts
```diff
@@ -67,8 +67,8 @@ const fields: ToApiConverters = {
   iWantToHavePapersServedAnotherWay: data => ({
     applicant1WantsToHavePapersServedAnotherWay: checkboxConverter(data.iWantToHavePapersServedAnotherWay),
   }),
-  legalProceedingsRelated: data => ({
-    legalProceedingsRelated: data.legalProceedings === YesOrNo.YES ? data.legalProceedingsRelated : [],
+  legalProceedingsByCase: data => ({
+    legalProceedingsByCase: data.legalProceedings === YesOrNo.YES ? data.legalProceedingsByCase : [],
```
velimd, line 71: as above, even if you change the name from the in the content, its really an array of legalProceedingsRelated so the return object is incorrect, itll have to be modified here or parser might need to be created where the data is converted.

PR #652, opened 2021-05-27
NFDIV-695 Add joint jurisdiction connection

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-695

### Change description ###
Added a jurisdiction connection for the joint journey

```
[ ] Yes
[x] No
```


src/main/app/case/definition.ts
```diff
@@ -374,6 +375,10 @@ export const enum JurisdictionConnections {
    * Applicant 2 is domiciled in England and Wales
    */
   APP_2_DOMICILED = 'I',
+  /**
+   * Applicant 1 habitually resides in England and Wales and joint application
+   */
+  APP_1_RESIDENT = 'J',
```
velimd, line 381:
```suggestion
  APP_1_RESIDENT_JOINT = 'J',
```

src/main/app/case/definition.ts
```diff
@@ -469,7 +474,7 @@ export const enum PaymentStatus {
   CANCELLED = 'cancelled',
   ERROR = 'error',
 }
-export const CASE_TYPE = 'NO_FAULT_DIVORCE12';
+export const CASE_TYPE = 'NO_FAULT_DIVORCE14';
```
velimd, line 477: this might break stuff as the LegalProcessed property has changed

velimd: so the test might never pass as 2 of the properties were changed, 1 was a minor changed but the other requires abit of thought.

src/main/app/case/definition.ts
lara-mpt: Yeah that's a better label, will need to update on backend first though

src/main/app/case/definition.ts
lara-mpt: Ah I needed to bump it to get the new jurisdiction field change

velimd: yhh, if you can convert it go for it. Maybe i missed something when looking at it. Tom has a branch you can work off or copy.

lara-mpt: cool, I'll look into it. Worst case could we revert his changes for the time being?

velimd: ideally, i would want that but then im not sure about the repercussion on the solicitor journey, we can also fake some of the data or pass null until its fixed but probably best to look at the ticket or discuss with Tom (hes away until tuesday)

velimd (approved):

PR #653, opened 2021-05-27
feat(NFDIV-680): Payment

SebFlippence:
### JIRA link ###

[NFDIV-680](https://tools.hmcts.net/jira/browse/NFDIV-680) Payment - State Changes

### Change description ###

Adds payment handling and triggers payment made event on success.

To test, use a [mock credit/debit card](https://docs.payments.service.gov.uk/testing_govuk_pay/#mock-card-numbers).

```
[ ] Yes
[x] No
```


src/main/app/case/CaseApi.ts
```diff
@@ -71,16 +83,23 @@ export class CaseApi {
     }
   }
 
-  public async triggerEvent(caseId: string, userData: Partial<Case>, eventName: string): Promise<CaseWithId> {
+  public async saveUserData(caseId: string, userData: Partial<Case>, eventName = CITIZEN_UPDATE): Promise<CaseWithId> {
```
linusnorton, line 86: nope

src/main/app/case/CaseApi.ts
```diff
@@ -28,12 +36,22 @@ export class CaseApi {
...
-        return { id: serviceCases[0].id, state: serviceCases[0].state, ...fromApiFormat(serviceCases[0].case_data) };
-      default:
+      }
+      case 1: {
+        const { id, state, case_data: caseData } = serviceCases[0];
+        return {
+          ...fromApiFormat(caseData),
+          id,
+          state,
+          applicationFeeOrderSummary: caseData.applicationFeeOrderSummary,
```
linusnorton, line 48: why are these two fields here? They're not special

src/main/app/case/CaseApi.ts
```diff
@@ -71,16 +89,29 @@ export class CaseApi {
...
+        event,
+        data,
+        event_token: token,
+      });
+
+      return {
+        ...fromApiFormat(caseData),
+        id,
+        state,
+        applicationFeeOrderSummary: caseData.applicationFeeOrderSummary,
```
linusnorton, line 112: just add them to the pass through

src/main/app/case/CaseApi.ts
```diff
@@ -71,16 +89,29 @@ export class CaseApi {
     }
   }
 
-  public async triggerEvent(caseId: string, userData: Partial<Case>, eventName: string): Promise<CaseWithId> {
+  public async saveUserData(caseId: string, userData: Partial<Case>, eventName = CITIZEN_UPDATE): Promise<CaseWithId> {
```
linusnorton, line 92: not really a fan of this, it doesn't seem like a good abstraction but happy to live with it if you are

src/main/app/controller/PaymentController.ts
```diff
@@ -0,0 +1,84 @@
...
+import { Response } from 'express';
+
+import { APPLICATION_SUBMITTED, HOME_URL, PAYMENT_CALLBACK_URL } from '../../steps/urls';
+import { CITIZEN_ADD_PAYMENT, CITIZEN_SUBMIT, PaymentStatus, State } from '../case/definition';
+import { Payment, PaymentClient } from '../payment/PaymentClient';
+import { PaymentModel } from '../payment/PaymentModel';
+
+import { AppRequest } from './AppRequest';
+
+export class PaymentController {
```
linusnorton, line 12: move this to the step as a default export and move the callback controller to a separate step

src/main/app/controller/PaymentController.ts
```diff
@@ -0,0 +1,84 @@
...
+    if (!payments.hasPayment) {
+      return res.redirect(HOME_URL);
+    }
+
+    const lastPaymentAttempt = payments.lastPayment;
+    let payment: Payment | undefined;
+    try {
+      payment = await paymentClient.get(lastPaymentAttempt.paymentReference);
+    } catch {
+      // ignore
```
linusnorton, line 63: maybe something a bit more robust here - set the status to failed with error of some sort?

src/main/app/controller/PaymentController.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    payments.setStatus(lastPaymentAttempt.paymentTransactionId, payment?.status);
+
+    req.session.userCase = await req.locals.api.triggerEvent(
+      req.session.userCase.id,
+      { payments: payments.list },
+      CITIZEN_ADD_PAYMENT
+    );
+
+    req.session.save(() => res.redirect(payments.wasLastPaymentSuccessful ? APPLICATION_SUBMITTED : HOME_URL));
```
linusnorton, line 74: is that HOME_URL a conscious choice? Maybe it should go to an error page saying the payment was not successful, please try again

src/main/modules/payment/index.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Adds payment middleware to add payment callbacks
+ */
+export class PaymentMiddleware {
+  public enableFor(app: Application): void {
+    const { errorHandler } = app.locals;
+
+    const paymentController = new PaymentController();
+
+    app.get(PAYMENT_URL, errorHandler(paymentController.payment.bind(paymentController)));
```
linusnorton, line 18: set these up with the rest of the controllers

linusnorton: in fact if you used a default export in the correct holder you wouldn't need to

src/main/modules/payment/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    app.get(PAYMENT_CALLBACK_URL, errorHandler(paymentController.callback.bind(paymentController)));
+
+    app.use(
+      errorHandler(async (req: AppRequest, res: Response, next: NextFunction) => {
+        if (
+          [State.Submitted, State.AwaitingDocuments, State.AwaitingHWFDecision].includes(req.session.userCase.state) &&
+          req.path !== APPLICATION_SUBMITTED
+        ) {
```
linusnorton, line 26: I think this can be in a more generic middleware like CaseStateMiddleware

src/main/app/case/CaseApi.ts
velimd: i agree

src/main/app/controller/PaymentController.ts
SebFlippence: Gov Pay handles this for us. I'll redirect to the check your answers pay.

Screenshot from 2021-06-04 12-35-36

src/main/app/controller/PaymentController.ts
SebFlippence: The error is logged in the payment client, and the `payments.setStatus` will set it to error if `payment` is not set, but I can remove the rethrowing of the error so that we don't need the try/catch here.

src/main/app/case/CaseApi.ts
SebFlippence: The main reason I separated these out was because it modifying the user data via `toApiFormat`, when this is moved to the PostController then I can revert this change.

src/main/app/case/CaseApi.ts
SebFlippence: I'll take a look at moving these into into `case.ts` and make sure that `toApiFormat` removes them if the user is sending the request.

src/main/app/case/CaseApi.ts
SebFlippence: The main reason I put these here rather than in the `case.ts` was for security.

Anything in `case.ts` can be modified by the user via a specially crafted post request.

src/main/app/case/CaseApi.ts
```diff
@@ -71,16 +83,29 @@ export class CaseApi {
     }
   }
 
-  public async triggerEvent(caseId: string, userData: Partial<Case>, eventName: string): Promise<CaseWithId> {
+  public async triggerEvent({
+    caseId,
+    data = {},
+    eventName = CITIZEN_UPDATE,
+    raw,
```
linusnorton, line 90: why not just pass it through as is in the toApiFormat?

src/main/app/case/CaseApi.ts
```diff
@@ -71,16 +83,29 @@ export class CaseApi {
     }
   }
 
-  public async triggerEvent(caseId: string, userData: Partial<Case>, eventName: string): Promise<CaseWithId> {
+  public async triggerEvent({
+    caseId,
+    data = {},
+    eventName = CITIZEN_UPDATE,
```
linusnorton, line 89: I don't think we should default value this. In think I need to understand why you're changing this method

src/main/modules/oidc/index.ts
```diff
@@ -43,5 +52,37 @@ export class OidcMiddleware {
         }
       })
     );
+
+    app.use(
+      errorHandler(async (req: AppRequest, res: Response, next: NextFunction) => {
+        if ([PAYMENT_URL, PAYMENT_CALLBACK_URL].includes(req.path as PageLink)) {
```
linusnorton, line 58: why put these in here?

src/main/app/case/from-api-format.ts
```diff
@@ -69,4 +76,7 @@ const fromApiDate = date => {
   return { year: `${+y}`, month: `${+m}`, day: `${+d}` };
 };
 
-export const fromApiFormat = (data: CaseData): Case => formatCase(fields, data);
+export const fromApiFormat = (data: CaseData): Case => ({
+  ...formatCase(fields, data),
+  ...formatCase(invert(readOnlyFormFieldsToCaseMapping), data),
```
velimd, line 81: 1. maybe you only need to call the method once with a new constructed object e.g. 
formatCase({ ...fields, ...invert(readOnlyFormFieldsToCaseMapping)}, data),
2. doesnt seem like theres a need to invert it but i guess for future proof?

src/main/app/case/CaseApi.ts
```diff
@@ -71,11 +85,10 @@ export class CaseApi {
     }
   }
 
-  public async triggerEvent(caseId: string, userData: Partial<Case>, eventName: string): Promise<CaseWithId> {
+  private async _triggerEvent(caseId: string, data: Partial<Case | CaseData>, eventName: string): Promise<CaseWithId> {
```
linusnorton, line 88: don't you dare start using _ for privates

src/main/app/check-your-answers/CheckYourAnswersPostController.ts
```diff
@@ -1,26 +1,26 @@
...
 
 @autobind
-export class InviteApplicant2PostController extends PostController<AnyObject> {
+export class CheckYourAnswersPostController extends PostController<AnyObject> {
   constructor(protected readonly form: Form) {
     super(form);
   }
 
-  public async post(req: AppRequest<AnyObject>, res: Response): Promise<void> {
+  public async post(req: AppRequest<AnyObject>, res: Response, next?: NextFunction): Promise<void> {
```
velimd, line 17: i think this all file can be simiplied similiar to how i do it here [link](https://github.com/hmcts/nfdiv-frontend/pull/697/files#diff-a725046819a0e0f28e62dd6ff56fd4ff7eea0abba5fcf80fb97220d93a120319R15)

src/main/app/payment/PaymentClient.ts
```diff
@@ -0,0 +1,160 @@
...
+  client: AxiosInstance;
+
+  constructor(private readonly session: AppSession, readonly returnUrl: string) {
+    this.client = Axios.create({
+      baseURL: config.get('services.payments.url'),
+      headers: {
+        Authorization: 'Bearer ' + session.user.accessToken,
+        ServiceAuthorization: getServiceAuthToken(),
+        'return-url': returnUrl,
+        'service-callback-url': '',
```
velimd, line 21: is this needed if we its blank?

src/main/steps/index.ts
```diff
@@ -76,7 +76,8 @@ export type StepWithContent = ({ generateContent: TranslationFn; form: FormConte
 export const stepsWithContent = ((): StepWithContent => {
   const results: StepWithContent = [];
   for (const step of sequence) {
-    const content = require(`${__dirname}${step.url}/content.ts`);
+    const stepContentFile = `${__dirname}${step.url}/content.ts`;
+    const content = fs.existsSync(stepContentFile) ? require(stepContentFile) : {};
```
velimd, line 80: is this check for the empty application submitted page?

src/main/app/case/from-api-format.ts
SebFlippence: :+1: Good spot, I've refactored to this and removed the unneeded `invert`

src/main/app/check-your-answers/CheckYourAnswersPostController.ts
SebFlippence: :+1: Nice, I've take your new implementation

src/main/app/payment/PaymentClient.ts
SebFlippence: :+1: Looks like it's optional, so I've removed it

src/main/steps/index.ts
SebFlippence: Yes, this is for steps that don't have content and handle logic only like the `/payment_callback` and the `/application-submitted` endpoints.

velimd (approved):

linusnorton (approved):

PR #686, opened 2021-06-03
NFDIV-696

tomxelliott:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-696

Screenshot 2021-06-03 at 17 26 45
Screenshot 2021-06-03 at 17 26 54


src/main/app/controller/PostController.ts
```diff
@@ -78,6 +78,29 @@ export class PostController<T extends AnyObject> {
     });
   }
 
+  public async sendToApplicant2ForReview(req: AppRequest<T>, res: Response, formData: Partial<Case>): Promise<void> {
```
SebFlippence, line 81: Would it still worked if you moved this method out of `PostController.ts` and into `InviteApplicant2PostController.ts`?

src/main/app/invite-applicant2/InviteApplicant2PostController.ts
```diff
@@ -0,0 +1,24 @@
...
+export class InviteApplicant2PostController extends PostController<AnyObject> {
+  constructor(protected readonly form: Form) {
+    super(form);
+  }
+
+  public async post(req: AppRequest<AnyObject>, res: Response): Promise<void> {
+    const { saveAndSignOut, saveBeforeSessionTimeout, _csrf, ...formData } = this.form.getParsedBody(req.body);
+
+    if (formData.applicationType === ApplicationType.JOINT_APPLICATION) {
+      await super.sendToApplicant2ForReview(req, res, formData);
```
SebFlippence, line 19: Then this would be

```suggestion
      await this.sendToApplicant2ForReview(req, res, formData);
```

tomxelliott: 👍

src/main/app/controller/PostController.ts
tomxelliott: 👍

tomxelliott: After trying it out unfortunately the suggestion didn't work due to this method using the private `save` method in the PostController.

src/main/app/controller/PostController.ts
```diff
@@ -78,6 +78,29 @@ export class PostController<T extends AnyObject> {
...
+    Object.assign(req.session.userCase, formData);
+    this.form.setFormState(req.session.userCase);
+    req.session.errors = this.form.getErrors(formData);
+
+    try {
+      const twoWeeks = 1000 * 60 * 60 * 24 * 14;
```
velimd, line 87:
```suggestion
      const twoWeeks = 14 * 24 * 60 * 60 * 1000;
```
just for constitency, not a big deal.

src/main/app/controller/PostController.ts
```diff
@@ -78,6 +78,29 @@ export class PostController<T extends AnyObject> {
...
+      formData.dueDate = new Date(Date.now() + twoWeeks);
```
linusnorton, line 88: I wonder if it would be best to set this field on the backend?

tomxelliott: Yeah I was weighing this up as well. Definitely happy to do it that way 👍

src/main/app/controller/PostController.ts
linusnorton: I think I agree with Seb, this is behaviour specific to a page and shouldn't be in the general controller. If you want to re-use the save method you could subclass and make it protected but it's possible you don't need that behaviour and just want to trigger the event

src/main/steps/sequence.ts
```diff
@@ -366,7 +366,12 @@ export const sequence: Step[] = [
   },
   {
     url: CHECK_ANSWERS_URL,
-    getNextStep: data => (data.applicant1HelpWithFeesRefNo ? APPLICATION_SUBMITTED : PAY_YOUR_FEE),
+    getNextStep: data =>
+      data.applicationType === ApplicationType.JOINT_APPLICATION
+        ? SENT_TO_APPLICANT2_FOR_REVIEW
+        : data.applicant1HelpWithFeesRefNo
+        ? APPLICATION_SUBMITTED
```
linusnorton, line 373: spacing is off here which makes this a bit hard to understand

src/main/app/invite-applicant2/InviteApplicant2PostController.ts
```diff
@@ -0,0 +1,24 @@
...
+@autobind
+export class InviteApplicant2PostController extends PostController<AnyObject> {
+  constructor(protected readonly form: Form) {
+    super(form);
+  }
+
+  public async post(req: AppRequest<AnyObject>, res: Response): Promise<void> {
+    const { saveAndSignOut, saveBeforeSessionTimeout, _csrf, ...formData } = this.form.getParsedBody(req.body);
+
+    if (formData.applicationType === ApplicationType.JOINT_APPLICATION) {
```
velimd, line 18: maybe we can have this in the actual post controller, does not seem like much is going on here so it can be another if like how we do save and signout or timeout?

src/main/steps/sequence.ts
tomxelliott: This was how the linter cleaned it up

linusnorton: ah the linter.

src/main/app/controller/PostController.ts
tomxelliott: sure

src/main/app/invite-applicant2/InviteApplicant2PostController.ts
tomxelliott: As I needed custom post behaviour, Seb said this was the way to do it

tomxelliott: I had previously had it the way you are describing and it wasn't working for me

SebFlippence: You could change the `save` method in the `PostController` to be `protected` and that way you could bring the new method from the `PostController` here.

SebFlippence (approved):

src/test/steps/check-your-answers.ts
```diff
@@ -24,3 +24,25 @@ Given("I've completed enough questions correctly to get to the check your answer
     ceremonyPlace: 'Maputo',
   });
 });
+
+Given("I've completed enough questions correctly to get to the check your answers page as a joint applicant", () => {
```
SebFlippence, line 28: Should this be `async`?

```suggestion
Given("I've completed enough questions correctly to get to the check your answers page as a joint applicant", async () => {
```

SebFlippence (approved):

PR #690, opened 2021-06-04
rename gender fields

linusnorton:


src/main/app/case/definition.ts
```diff
@@ -93,6 +93,7 @@ export interface Applicant {
   HomeAddress: AddressGlobalUK;
   PhoneNumber: string;
   ContactDetailsConfidential: ConfidentialAddress;
+  Gender: Gender;
```
velimd, line 96: didnt we already have this?

src/main/app/case/definition.ts
```diff
@@ -366,7 +367,7 @@ export const enum Gender {
  * - `G` - Eligible for Residual Jurisdiction
  * - `H` - Applicant 1 is domiciled in England and Wales
  * - `I` - Applicant 2 is domiciled in England and Wales
- * - `J` - Applicant 1 habitually resides in England and Wales and joint application
+ * - `J` - Applicant 1 habitually resides in England and Wales
```
velimd, line 370: was this not a requirement?

src/main/app/case/definition.ts
```diff
@@ -406,7 +407,7 @@ export const enum JurisdictionConnections {
    */
   APP_2_DOMICILED = 'I',
   /**
-   * Applicant 1 habitually resides in England and Wales and joint application
+   * Applicant 1 habitually resides in England and Wales
    */
   APP_1_RESIDENT_JOINT = 'J',
```
velimd, line 412: if its not joint, maybe remove JOINT part

velimd (approved):

PR #711, opened 2021-06-10
NFDIV-743

tomxelliott:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-743

Screenshot 2021-06-10 at 14 04 58


src/main/app/access-code/AccessCodePostController.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { getNextStepUrl } from '../../steps';
+import { getCaseApi } from '../case/CaseApi';
+import { CITIZEN_LINK_APPLICANT_2 } from '../case/definition';
+import { AppRequest } from '../controller/AppRequest';
+import { AnyObject, PostController } from '../controller/PostController';
+import { Form } from '../form/Form';
+
+@autobind
+export class AccessCodePostController extends PostController<AnyObject> {
```
linusnorton, line 12: I don't think you need to extend PostController really

src/main/app/access-code/AccessCodePostController.ts
```diff
@@ -0,0 +1,54 @@
...
+  constructor(protected readonly form: Form) {
+    super(form);
+  }
+
+  public async post(req: AppRequest<AnyObject>, res: Response): Promise<void> {
+    req.locals.api = getCaseApi(req.session.user, req.locals.logger);
```
linusnorton, line 18: don't you need to do this as a caseworker? The user won't have access yet

src/main/app/access-code/AccessCodePostController.ts
tomxelliott: it was so I could access the protected save method

src/main/app/access-code/AccessCodePostController.ts
tomxelliott: I make the call in the case-api with the caseworker, but I need this user token so I know who to link the case role to

src/main/steps/sequence.ts
```diff
@@ -385,4 +387,8 @@ export const sequence: Step[] = [
     url: JURISDICTION_CONNECTION_SUMMARY,
     getNextStep: () => YOUR_NAME,
   },
+  {
```
linusnorton, line 390: I don't think these pages should go into the steps array, that's just for applicant 1

tomxelliott: Where would be the best place for them?

src/main/app/access-code/AccessCodePostController.ts
linusnorton, line 12: bonus points for putting this file in the correct folder (and not the controller folder)

src/main/app/access-code/AccessCodePostController.ts
```diff
@@ -0,0 +1,54 @@
...
+    const { saveAndSignOut, saveBeforeSessionTimeout, _csrf, ...formData } = this.form.getParsedBody(req.body);
+
+    Object.assign(req.session.userCase, formData);
+    this.form.setFormState(req.session.userCase);
+    req.session.errors = this.form.getErrors(formData);
+
+    try {
+      const caseData = await req.locals.api.getCaseById(formData.caseReference as string);
+
+      if (caseData.invitePin !== formData.accessCode) {
```
linusnorton, line 29: I think this check should be done in the API callback

src/main/app/access-code/AccessCodePostController.ts
tomxelliott: At this point the respondent would have created their idam user profile and been redirected to the enter-your-access-code page as we would have that as the redirect link as part of the link in the email they receive.

config/default.yaml
```diff
@@ -12,6 +12,8 @@ services:
     tokenURL: 'https://idam-api.aat.platform.hmcts.net/o/token'
     clientID: 'divorce'
     clientSecret: 'NEED TO INSERT SECRET'
+    caseworkerUsername: 'DivCaseWorkerUser@AAT.com'
+    caseworkerPassword: 'DivPassword1234'
```
linusnorton, line 16: take these out, there is something to load them from the vault

src/main/app/access-code/AccessCodePostController.ts
```diff
@@ -0,0 +1,57 @@
...
+
+@autobind
+export class AccessCodePostController extends PostController<AnyObject> {
+  constructor(protected readonly form: Form) {
+    super(form);
+  }
+
+  public async post(req: AppRequest<AnyObject>, res: Response): Promise<void> {
+    const caseworkerUser = await getCaseWorkerUser();
+    req.locals.api = getCaseApi(caseworkerUser, req.locals.logger);
```
linusnorton, line 20: My preference would not to be set this in locals as it's hard to know what else uses it and we want to make absolutely sure this API is only used to load and link the case. It would require changing `this.save(...)` below to call the event in the API manually, but I don't know why you haven't done that anyway - is there something in `this.save(...)` you need?

tomxelliott: this.save() is just the save in the PostController

linusnorton: yeah that's what I mean - is there a reason you need to use that, you could just trigger the event on the api

src/main/steps/urls.ts
```diff
@@ -72,6 +72,9 @@ export const DOCUMENT_MANAGER: PageLink = '/document-manager';
 export const SENT_TO_APPLICANT2_FOR_REVIEW: PageLink = '/application-sent-for-review';
 export const PAY_YOUR_FEE: PageLink = '/pay-your-fee';
 
+export const ENTER_YOUR_ACCESS_CODE: PageLink = '/enter-your-access-code';
```
SebFlippence, line 75: Does there need to be anything in the `src/main/modules/state-redirect/index.ts` to make sure that applicant 2 is redirected to this page when they register/login for the first time or is this taken care of by the email?

tomxelliott: @lara-mpt is currently working on the link that the second applicant will use and that will redirect them to the enter-your-access-code page

SebFlippence (approved):

linusnorton (approved):

PR #714, opened 2021-06-11
fix: nightly cross browser tests

SebFlippence:
### JIRA link ###

N/A

### Change description ###

Some nightly cross browser tests are silently failing, this PR should fix them.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/cross-browser/codecept.conf.ts
+++ b/src/test/cross-browser/codecept.conf.ts
@@ -4,13 +4,12 @@ import { config as testConfig } from '../config';
 
 setHeadlessWhen(testConfig.TestHeadlessBrowser);
 
-const url = testConfig.TEST_URL || 'http://localhost:3001';
 let helpers = {};
 let plugins = {};
 if (process.env.SAUCE === 'true') {
   helpers = {
     WebDriver: {
-      url,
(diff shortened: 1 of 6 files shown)


src/test/cross-browser/codecept.conf.ts
```diff
@@ -4,13 +4,12 @@ import { config as testConfig } from '../config';
...
 setHeadlessWhen(testConfig.TestHeadlessBrowser);
 
-const url = testConfig.TEST_URL || 'http://localhost:3001';
 let helpers = {};
 let plugins = {};
 if (process.env.SAUCE === 'true') {
   helpers = {
     WebDriver: {
-      url,
+      url: testConfig.TEST_URL,
```
velimd, line 12: should we not have an or condition like above?

velimd (approved):

src/test/cross-browser/codecept.conf.ts
SebFlippence: I've moved it up into the main config so `testConfig.TEST_URL` still doing `process.env.TEST_URL || 'http://localhost:3001'` 👍

PR #717, opened 2021-06-11
feat(NFDIV-802): Allow address entry with no JS

SebFlippence:
### JIRA link ###

[NFDIV-802](https://tools.hmcts.net/jira/browse/NFDIV-802) Allow postal address entry when JavaScript is disabled

### Change description ###

Allows postal address entry when JavaScript is disabled.

```
[ ] Yes
[x] No
```


src/main/steps/enter-your-address/get.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@autobind
+export default class YourAddressGetController extends GetController {
+  constructor() {
+    super(__dirname + '/template', generateContent);
+  }
+
+  public async get(req: AppRequest, res: Response): Promise<void> {
+    if (Object.keys(req.query).includes('postcode')) {
+      req.session.userCase.applicant1AddressPostcode = req.query.postcode?.toString();
```
linusnorton, line 17: is there any way to avoid this? otherwise we'll need a custom controller for each time we ask for an address

SebFlippence: 👍 I've removed these additional get controllers and I hiden the postcode lookup if JavaScript is disabled

linusnorton (approved):

PR #746, opened 2021-06-21
NFDIV-752: Create get ccd case role endpoint client

tomxelliott:


src/main/app/case/CaseApi.ts
```diff
@@ -95,6 +96,16 @@ export class CaseApi {
     }
   }
 
+  public async getCaseUserRoles(caseId: string, userId: string): Promise<CaseAssignedUserRoles> {
```
velimd, line 99: should it not be called after the user logs in?

velimd (approved):

src/main/app/case/CaseApi.ts
tomxelliott: Yeah that is the plan 👍

SebFlippence (approved):

PR #731, opened 2021-06-16
feat(NFDIV-749): Added applicant 2 enter your name

SebFlippence:
### JIRA link ###

[NFDIV-749](https://tools.hmcts.net/jira/browse/NFDIV-749) Enter your name - Joint App 2

### Change description ###

This reuses the existing step/question for applicant 2 first, middle and last names for applicant 2.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #760, opened 2021-06-23
Folder restructing + NFDIV-762

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-762

### Change description ###
restructing folder steps for applicant1
implementing https://tools.hmcts.net/jira/browse/NFDIV-762

```
[ ] Yes
[X] No
```


src/main/steps/applicant2Sequence.ts
```diff
@@ -1,13 +1,39 @@
...
-export const applicant2Sequence: Step[] = [
+const sequences: Step[] = [
   {
     url: YOU_NEED_TO_REVIEW_YOUR_APPLICATION,
-    getNextStep: () => HAS_RELATIONSHIP_BROKEN_APPLICANT2,
+    getNextStep: () => HAS_RELATIONSHIP_BROKEN_URL,
+  },
+  {
+    url: HAS_RELATIONSHIP_BROKEN_URL,
+    // TODO - to be replaced with Has your marriage irretrievably broken down page once developed
```
SebFlippence, line 20: Looks like this `TODO` is done? 🎉

velimd: What is the next page for this page?

SebFlippence: For this "relationship broken down" page AFAIK it goes to the "enter your name" page, which I have a PR https://github.com/hmcts/nfdiv-frontend/pull/731 for that adds this to this file. The Miro board should have the full sequence.

But this comment looks like it should be removed since its saying add the broken down page here, which has been added as `HAS_RELATIONSHIP_BROKEN_APPLICANT2`.

linusnorton (commented):
I'm not sure I would take the approach of prefixing ${APPLICANT2} to the URLs and re-using the applicant 1 URL strings but it's not a big deal

velimd:
> I'm not sure I would take the approach of prefixing ${APPLICANT2} to the URLs and re-using the applicant 1 URL strings but it's not a big deal

i thought it would be better than manually adding it to every applicant2 urls

src/main/steps/applicant2Sequence.ts
velimd: yhh, ill fix the comment and as its in your pr, ill leave the nextpage.

linusnorton:
> > I'm not sure I would take the approach of prefixing ${APPLICANT2} to the URLs and re-using the applicant 1 URL strings but it's not a big deal
> 
> i thought it would be better than manually adding it to every applicant2 urls

I see the logic and you're probably right. My only thought was I wanted to decouple the journeys as much as possible but that leads to a bit duplication. Swings and roundabouts

velimd:
> > > I'm not sure I would take the approach of prefixing ${APPLICANT2} to the URLs and re-using the applicant 1 URL strings but it's not a big deal
> > 
> > 
> > i thought it would be better than manually adding it to every applicant2 urls
> 
> I see the logic and you're probably right. My only thought was I wanted to decouple the journeys as much as possible but that leads to a bit duplication. Swings and roundabouts

fair, im good either way. with manually adding it, we dont need the function and itll make the sequence files consistent so i dont mind.

linusnorton:
> > > > I'm not sure I would take the approach of prefixing ${APPLICANT2} to the URLs and re-using the applicant 1 URL strings but it's not a big deal
> > > 
> > > 
> > > i thought it would be better than manually adding it to every applicant2 urls
> > 
> > 
> > I see the logic and you're probably right. My only thought was I wanted to decouple the journeys as much as possible but that leads to a bit duplication. Swings and roundabouts
> 
> fair, im good either way. with manually adding it, we dont need the function and itll make the sequence files consistent so i dont mind.

keep it as is

SebFlippence (approved):

PR #765, opened 2021-06-24
NFDIV-755 & NFDIV-778

tomxelliott:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-755
https://tools.hmcts.net/jira/browse/NFDIV-778


src/main/steps/how-do-you-want-to-apply/post.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  public async post(req: AppRequest<AnyObject>, res: Response, next?: NextFunction): Promise<void> {
+    const { saveAndSignOut, saveBeforeSessionTimeout, _csrf, ...formData } = this.form.getParsedBody(req.body);
+
+    if (
+      formData.applicationType === ApplicationType.SOLE_APPLICATION &&
+      req.session.userCase.applicant2AddressPrivate === YesOrNo.YES
+    ) {
+      req.body.applicant2Address1 = null;
```
velimd, line 22: not sure why but i feel like this is not needed (due to getUnreachableAnswers) or could be moved to the to-api-format check.

velimd: not a big deal but im getting that feeling.

tomxelliott: This was A/C 5 on NFDIV-778 - happy to move this if it isn't required

src/main/steps/applicant2Sequence.ts
```diff
@@ -10,4 +16,13 @@ export const applicant2Sequence: Step[] = [
...
     getNextStep: () => HOME_URL,
   },
+  {
+    url: HOW_THE_COURTS_WILL_CONTACT_YOU_APPLICANT2,
+    getNextStep: () => ADDRESS_PRIVATE_APPLICANT2,
+  },
+  {
+    url: ADDRESS_PRIVATE_APPLICANT2,
+    // TODO - to be replaced by Enter your postal address when developed as part of NFDIV-757
+    getNextStep: () => HOME_URL,
```
SebFlippence, line 26: I think it probably best to link it to the page it should go to. This way it will satisfy the ACs (even if it shows a 404) and we can have a functional test that checks goes to the correct page e.g. `Then the page URL should be "/applicant2/enter-your-address"`.

```suggestion
    getNextStep: () => YOUR_NAME,
```

velimd (approved):

PR #781, opened 2021-06-28
NFDIV-689

tomxelliott:
Additional userCase object that could be null when logging in as App 2

### DIFF
--- a/src/main/app/controller/GetController.ts
+++ b/src/main/app/controller/GetController.ts
@@ -44,7 +44,7 @@ export class GetController {
       ...content,
       sessionErrors,
       htmlLang: language,
-      isDraft: req.session?.userCase.state ? req.session.userCase.state === State.Draft : true,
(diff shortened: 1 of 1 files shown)


velimd (approved):

src/main/app/controller/GetController.ts
```diff
@@ -44,7 +44,7 @@ export class GetController {
       ...content,
       sessionErrors,
       htmlLang: language,
-      isDraft: req.session?.userCase.state ? req.session.userCase.state === State.Draft : true,
+      isDraft: req.session?.userCase?.state ? req.session.userCase.state === State.Draft : true,
```
SebFlippence, line 47: Might be worth updating the type for this in `src/main/app/controller/AppRequest.ts` to be `userCase?: CaseWithId;`. This way TypeScript will highlight any other places were it might not be set.

tomxelliott: Was throwing a lot of errors mainly to do with the tests - will investigate separately so that I don't hold up merging this fix in.

PR #798, opened 2021-06-30
NFDIV-976: Trigger app2 update if the user logged in isApplicant2

tomxelliott:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-976

### DIFF
--- a/src/main/app/controller/PostController.ts
+++ b/src/main/app/controller/PostController.ts
@@ -37,7 +37,7 @@ export class PostController<T extends AnyObject> {
       await this.saveAndSignOut(req, res, stepData);
     } else if (req.body.saveBeforeSessionTimeout) {
       await this.saveBeforeSessionTimeout(req, res, stepData);
-    } else if (req.body.isApplicant2) {
+    } else if (req.session.isApplicant2) {
       await this.saveAndContinue(req, res, stepData, CITIZEN_APPLICANT2_UPDATE);
     } else {
       await this.saveAndContinue(req, res, stepData, eventName);
--- a/src/main/app/controller/PostController.ts
+++ b/src/main/app/controller/PostController.ts
@@ -21,7 +21,7 @@ export class PostController<T extends AnyObject> {
     req: AppRequest<T>,
     res: Response,
     next?: NextFunction,
-    eventName: string = CITIZEN_UPDATE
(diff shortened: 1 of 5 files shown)


src/main/app/controller/PostController.ts
```diff
@@ -37,6 +37,8 @@ export class PostController<T extends AnyObject> {
       await this.saveAndSignOut(req, res, stepData);
     } else if (req.body.saveBeforeSessionTimeout) {
       await this.saveBeforeSessionTimeout(req, res, stepData);
+    } else if (req.body.isApplicant2) {
+      await this.saveAndContinue(req, res, stepData, CITIZEN_APPLICANT2_UPDATE);
```
lara-mpt, line 41: The CITIZEN_UPDATE event is being passed as 'eventName', while this is just CITIZEN_APPLICANT2_UPDATE
  Could this be more consistent maybe?

tomxelliott: Not sure what you mean?

lara-mpt: On line 43:
      await this.saveAndContinue(req, res, stepData, eventName);
      
Where 'eventName' is CITIZEN_UPDATE. 
Just think this could be a bit confusing

lara-mpt (approved):

src/main/app/controller/PostController.ts
```diff
@@ -37,6 +37,8 @@ export class PostController<T extends AnyObject> {
       await this.saveAndSignOut(req, res, stepData);
     } else if (req.body.saveBeforeSessionTimeout) {
       await this.saveBeforeSessionTimeout(req, res, stepData);
+    } else if (req.body.isApplicant2) {
```
SebFlippence, line 40: Should this be 
```suggestion
    } else if (req.session.isApplicant2) {
```

src/main/app/controller/PostController.ts
```diff
@@ -37,6 +37,8 @@ export class PostController<T extends AnyObject> {
       await this.saveAndSignOut(req, res, stepData);
     } else if (req.body.saveBeforeSessionTimeout) {
       await this.saveBeforeSessionTimeout(req, res, stepData);
+    } else if (req.session.isApplicant2) {
```
velimd, line 40: i think you can have an if statement before this with this check but just sets eventName to CITIZEN_APPLICANT2_UPDATE so you dont need another if statement here to call the same method as below.

velimd (approved):

PR #811, opened 2021-07-01
Content when logged in as Applicant 2 was being rendered as if Applicant 1

tomxelliott:
Update common content to check whether the user is applicant 2, if they are, we return the correct content for husband vs wife.

### DIFF
--- a/src/main/steps/common/common.content.ts
+++ b/src/main/steps/common/common.content.ts
@@ -160,17 +160,19 @@ export const generatePageContent = ({
   isDivorce = true,
   formState,
   userEmail,
+  isApplicant2,
 }: {
   language: Language;
   pageContent?: TranslationFn;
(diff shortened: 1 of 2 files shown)


src/main/steps/common/common.content.ts
```diff
@@ -199,15 +201,20 @@ const getServiceName = (translations: typeof en, isDivorce: boolean): string =>
...
+  selectedGender: Gender,
+  isDivorce: boolean,
+  isApplicant2: boolean
+): string => {
   if (!isDivorce) {
     return translations.civilPartner;
   }
   if (selectedGender === Gender.MALE) {
-    return translations.husband;
+    return isApplicant2 ? translations.wife : translations.husband;
```
lara-mpt, line 214: Not sure if this will take into account same sex marriages

PR #857, opened 2021-07-08
NFDIV-778

tomxelliott:
Ensure keep address private field is not incorrectly pre-populated

Screenshot 2021-07-08 at 15 52 14

### DIFF
--- a/src/main/app/case/to-api-format.ts
+++ b/src/main/app/case/to-api-format.ts
@@ -61,7 +61,11 @@ const fields: ToApiConverters = {
   }),
   applicant2AddressPrivate: data => ({
     applicant2ContactDetailsConfidential:
-      data.applicant2AddressPrivate === YesOrNo.YES ? ConfidentialAddress.KEEP : ConfidentialAddress.SHARE,
+      data.applicant2AddressPrivate === YesOrNo.YES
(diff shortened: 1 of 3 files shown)


src/main/app/case/to-api-format.ts
```diff
@@ -61,7 +61,11 @@ const fields: ToApiConverters = {
   }),
   applicant2AddressPrivate: data => ({
     applicant2ContactDetailsConfidential:
-      data.applicant2AddressPrivate === YesOrNo.YES ? ConfidentialAddress.KEEP : ConfidentialAddress.SHARE,
+      data.applicant2AddressPrivate === YesOrNo.YES
+        ? ConfidentialAddress.KEEP
+        : data.applicant2AddressPrivate === null
```
velimd, line 66: i dont see why we are checking for null?

velimd (approved):

src/main/app/case/to-api-format.ts
tomxelliott: As discussed, the app 2 questions are all pulled through as null once app 2 is linked but because there is a handler for this particular field the field is getting set to No.

velimd (approved):

PR #876, opened 2021-07-13
Configure getAllPossibleAnswers to include Applicant2 Sequence when application type is Joint

tomxelliott:


src/main/app/case/answers/possibleAnswers.ts
```diff
@@ -19,6 +19,8 @@ const IGNORE_UNREACHABLE_FIELDS = [
   'payments',
 ];
 
+const applicant2SequenceStart = '/applicant2/you-need-to-review-your-application';
```
linusnorton, line 22: shouldn't you use the const in the url file?

tomxelliott: yeah, good point!

src/main/app/case/answers/possibleAnswers.ts
```diff
@@ -37,7 +39,14 @@ export const getAllPossibleAnswers = (caseState: Partial<Case>, steps: Step[]):
     return fields;
   };
 
-  return getPossibleFields(sequenceWithForms[0]);
+  const applicant1Fields = getPossibleFields(sequenceWithForms[0]);
```
velimd, line 42: i feel like you could reduce this and basically pass the individual sequences to the function.

velimd:
```suggestion
  const possibleAnswers = [...getAllPossibleAnswers(userCase, stepsWithContentApplicant1), ...getAllPossibleAnswers(userCase, stepsWithContentApplicant2)];
```
couldnt edit where it should go but i thought where the function is called.

tomxelliott: I see what you mean, will just need to control when app2 one is included so that it is only called when the application is joint.

src/main/app/case/answers/possibleAnswers.ts
```diff
@@ -44,7 +45,19 @@ export const omitUnreachableAnswers = (caseState: Partial<Case>, steps: Step[]):
   pick(caseState, getAllPossibleAnswers(caseState, steps));
 
 export const getUnreachableAnswersAsNull = (userCase: Partial<Case>): Partial<Case> => {
-  const possibleAnswers = getAllPossibleAnswers(userCase, stepsWithContent);
+  const possibleAnswers = [
+    ...getAllPossibleAnswers(userCase, stepsWithContent),
```
velimd, line 49:
```suggestion
    ...getAllPossibleAnswers(userCase, stepsWithContentApplicant1),
```

src/main/app/case/answers/possibleAnswers.ts
```diff
@@ -44,7 +45,19 @@ export const omitUnreachableAnswers = (caseState: Partial<Case>, steps: Step[]):
...
+    ...(userCase.applicationType === ApplicationType.JOINT_APPLICATION
+      ? getAllPossibleAnswers(
+          userCase,
+          stepsWithContent.slice(
```
velimd, line 53:
```suggestion
      getAllPossibleAnswers(userCase, stepsWithContentApplicant2),
```

velimd: would this not work?

tomxelliott: As discussed, we cannot make this change until Seb's branch is merged in at a later date. This can be done as some refactoring once this happens.

src/main/app/case/answers/possibleAnswers.ts
tomxelliott: To be refactored once Seb's branch is merged

velimd (approved):

PR #890, opened 2021-07-16
attempt to fix cross browser tests

linusnorton:


src/test/config.ts
```diff
@@ -28,7 +28,7 @@ lockFile.unlockSync(lock);
 
 const propertiesVolume = new PropertiesVolume();
 propertiesVolume.enableFor({
-  locals: { developmentMode: !process.env.JENKINS_HOME },
+  locals: { developmentMode: !process.env.CI },
```
velimd, line 31: where does this get set?

linusnorton: its a more generic check that the script is being run on a CI server. AFAIK jenkins, github actions, travis all set it

velimd (approved):

PR #906, opened 2021-07-20
Nfdiv 722 confirm your joint application page

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-722

### Change description ###
Add the Confirm your joint application page

```
[ ] Yes
[x] No
```
Screenshot 2021-08-09 at 11 27 52
Screenshot 2021-08-09 at 11 28 00
Screenshot 2021-08-09 at 11 28 11
Screenshot 2021-08-09 at 11 28 21


src/main/steps/applicant1/confirm-your-joint-application/content.ts
```diff
@@ -0,0 +1,206 @@
...
+import { Checkbox } from '../../../app/case/case';
+import { YesOrNo } from '../../../app/case/definition';
+import { TranslationFn } from '../../../app/controller/GetController';
+import { FormContent } from '../../../app/form/Form';
+import { isFieldFilledIn } from '../../../app/form/validation';
+import { Sections } from '../../applicant1Sequence';
+import { CommonContent } from '../../common/common.content';
+import * as urls from '../../urls';
+
+const en = ({ isDivorce, partner, formState }: CommonContent) => ({
```
velimd, line 12: is alot of these duplicates from app1 and 2 CYA? if so maybe we can import it from them and add extra stuff we need.

src/main/app/case/answers/getAnswerRows.ts
```diff
@@ -27,16 +28,15 @@ export const getAnswerRows = function (
...
+  const { stepsWithContent, processedFormState } = setUpSteps(
+    formState,
+    isCompleteCase,
+    overrideStepsContent,
+    isApplicant2
+  );
 
   return stepsWithContent
-    .filter(step => step.showInSection === section)
+    .filter(step => (isCompleteCase ? step.showInCompleteSection === section : step.showInSection === section))
```
velimd, line 39: as alot of these are duplicates, maybe we can do
```suggestion
    .filter(step => (isCompleteCase && step.showInCompleteSection ? step.showInCompleteSection === section : step.showInSection === section))
```
that way we dont need to duplicate it in the sequence.

src/main/app/case/answers/getAnswerRows.ts
```diff
@@ -83,6 +83,18 @@ export const getAnswerRows = function (
               }),
         });
 
+      const addCompleteQuestionAnswer = (question: string, answer: string) => {
```
velimd, line 86: why not use the one above and set showActions to false?

src/main/app/case/answers/getAnswerRows.ts
```diff
@@ -111,12 +123,75 @@ export const getAnswerRows = function (
         }
         const customAnswerWithHtml = this.ctx.stepAnswersWithHTML?.[step.url]?.[fieldKey];
 
-        addQuestionAnswer(
-          customQuestion || (question as string),
-          this.env.filters.nl2br(this.env.filters.escape(customAnswer ?? answer)),
-          this.ctx.stepLinks[step.url],
-          customAnswerWithHtml
-        );
+        if (isCompleteCase) {
```
velimd, line 126: as mentioned above, if you use addQuestionAnswer, you dont need this condition.

src/main/app/case/answers/getAnswerRows.ts
```diff
@@ -111,12 +123,75 @@ export const getAnswerRows = function (
...
+            customQuestion || (question as string),
+            this.env.filters.nl2br(this.env.filters.escape(customAnswer ?? answer)),
+            this.ctx.stepLinks[step.url],
+            customAnswerWithHtml
+          );
+        }
+      }
+
+      if (isCompleteCase) {
+        if (section === 'aboutApplicant1' && step.url === YOUR_NAME) {
```
velimd, line 142: wouldnt these be set already due to being a content page with form?

src/main/steps/applicant1Sequence.ts
```diff
@@ -92,6 +96,7 @@ export const applicant1Sequence: Step[] = [
   {
     url: HAS_RELATIONSHIP_BROKEN_URL,
     showInSection: Sections.AboutPartnership,
+    showInCompleteSection: Sections.AboutPartnership,
```
velimd, line 99: as mentioned above, with the condition i suggested, we dont need to duplicate these, just ones that are different.

src/main/app/case/answers/getAnswerRows.ts
lara-mpt: @velimd This doesn't work because if it is a Complete section, and we don't want a specific step to show up, but the step is assigned a 'showInSection' it will show up

src/main/steps/applicant1/confirm-your-joint-application/template.njk
```diff
@@ -0,0 +1,37 @@
...
+        classes: 'govuk-!-margin-bottom-9',
+        rows: rows
+      }) }}
+    {% endif %}
+  {% endfor %}
+
+  {% block form %}
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        {% include "./form.njk" %}
```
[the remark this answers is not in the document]
lara-mpt: This template has a few differences, so would be best to keep them separate I think

src/main/steps/applicant1/confirm-your-joint-application/content.ts
lara-mpt: There's only a small overlap - 1 question and 3 answers, I'm not sure it would be worth it

src/main/app/case/answers/getAnswerRows.ts
```diff
@@ -62,7 +62,7 @@ export const getAnswerRows = function (
       const addQuestionAnswer = (question: string, answer: string, link?: PageLink, html?: string) =>
         questionAnswers.push({
           key: {
-            text: question,
+            html: question,
```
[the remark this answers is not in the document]
lara-mpt: Tried it the way you suggested, and it doesn't work, I don't think you can escape it as text

src/main/steps/applicant1Sequence.ts
lara-mpt: There are some steps which have Sections assigned that we don't want to show up in the completeSections, so I think we'll need this to differentiate between them

src/main/app/case/answers/getAnswerRows.ts
```diff
@@ -110,15 +110,72 @@ export const getAnswerRows = function (
           continue;
         }
         const customAnswerWithHtml = this.ctx.stepAnswersWithHTML?.[step.url]?.[fieldKey];
+        const stepLinks = isCompleteCase ? undefined : this.ctx.stepLinks[step.url];
```
velimd, line 113: is this needed? because showActions is set to false, so the links shouldnt appear.

src/main/steps/applicant1/confirm-your-joint-application/template.njk
```diff
@@ -0,0 +1,37 @@
...
+{% block content %}
+  {% block error_summary %}
+    {% include "common/error/summary.njk" %}
+  {% endblock %}
+
+  <h1 class="govuk-heading-xl">{{ title }}</h1>
+  <h3 class="govuk-body-l">{{ subHeader }}</h3>
+
+  {% for key, section in sections %}
+    {% set rows = getAnswerRows(section, true, false) %}
```
velimd, line 17: maybe best to set this as variables.

src/main/app/case/answers/getAnswerRows.ts
lara-mpt: Yeah when I tried using just this.ctx.stepLinks[step.url]; it would error out

src/main/steps/applicant2/confirm-your-joint-application/post.ts
```diff
@@ -0,0 +1,19 @@
...
+
+@autobind
+export default class ConfirmYourJointApplicationPostController extends PostController<AnyObject> {
+  constructor(protected readonly form: Form) {
+    super(form);
+  }
+
+  public async post(req: AppRequest<AnyObject>, res: Response, next?: NextFunction): Promise<void> {
+    const event = APPLICANT_2_APPROVE;
+    await super.post(req, res, next, event);
```
tomxelliott, line 17: probably no need to create the const `event`. just pass the APPLICANT_2_APPROVE directly to the method?

tomxelliott:
```suggestion
    await super.post(req, res, next, APPLICANT_2_APPROVE);
```

tomxelliott (approved):

PR #915, opened 2021-07-21
Use preview appinsights

linusnorton:


charts/nfdiv-frontend/values.preview.template.yaml
```diff
@@ -8,6 +8,11 @@ nodejs:
     primaryIngressHost: ${SERVICE_FQDN}
   autoscaling:
     enabled: false
+  keyVaults:
+    nfdiv:
+      resourceGroup: nfdiv
+      secrets:
+        - AppInsightsInstrumentationKey-Preview
```
nitinprabhu, line 15: do you not require other secrets here ? https://github.com/hmcts/nfdiv-frontend/blob/7467555d537b05d5809f3bc98b6833b2dcaa6ed7/charts/nfdiv-frontend/values.yaml#L27-L31

charts/nfdiv-frontend/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: '1.0'
 description: A Helm chart for nfdiv-frontend App
 name: nfdiv-frontend
 home: https://github.com/hmcts/nfdiv-frontend
-version: 0.0.39
+version: 0.0.40
```
nitinprabhu, line 6: don't think we need to bump chart version

nitinprabhu (approved):

PR #951, opened 2021-07-29
Nfdiv 723

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-723

### Change description ###

```
[ ] Yes
[X] No
```


src/test/functional/features/pay-and-submit.feature
```diff
@@ -0,0 +1,29 @@
+# TODO uncomment when joint application can be submitted
```
tomxelliott, line 1: When will this be re-added?

velimd: i guess when the full app2 journey is implemented as i cannot get it in state. (tho i think Lara is working on it).

PR #961, opened 2021-08-02
Update payment client

linusnorton:
- Update payment client to use case_type rather than site_id
- Use new payments model
- Remove unnecessary code added to deal with API adding spurious payment record

### DIFF
--- a/src/main/app/case/from-api-format.ts
+++ b/src/main/app/case/from-api-format.ts
@@ -31,6 +31,7 @@ const checkboxConverter = (value: string | undefined) => {
 
 const fields: FromApiConverters = {
   ...invert(formFieldsToCaseMapping),
+  ...invert(readOnlyFormFieldsToCaseMapping),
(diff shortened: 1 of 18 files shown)


src/main/app/case/from-api-format.ts
```diff
@@ -31,6 +31,7 @@ const checkboxConverter = (value: string | undefined) => {
 
 const fields: FromApiConverters = {
   ...invert(formFieldsToCaseMapping),
+  ...invert(readOnlyFormFieldsToCaseMapping),
```
velimd, line 34: My memory is bad but i think there was a reason this was removed (i believe it was here before).

tomxelliott (approved):

velimd (approved):

src/main/app/case/from-api-format.ts
linusnorton: well, lets find out

PR #999, opened 2021-08-10
Clean up post controllers

linusnorton:
- Remove the event name from the post method in favour of a protected method to get the correct event name.
- Fix bug in CYA form validation
- Change the way the prayer gets unset to prevent app2 from unsetting it
- Fix local cucumber tests


src/main/steps/applicant1/check-your-answers/get.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Response } from 'express';
+
+import { Checkbox } from '../../../app/case/case';
+import { AppRequest } from '../../../app/controller/AppRequest';
+import { GetController } from '../../../app/controller/GetController';
+
+@autobind
+export default class CheckYourAnswersGetController extends GetController {
+  public async get(req: AppRequest, res: Response): Promise<void> {
+    req.session.userCase.applicant1IBelieveApplicationIsTrue = Checkbox.Unchecked;
```
velimd, line 11: this is abit tricky, even tho you unchecked it here, itll still be checked in the casedata.

velimd (commented):
other than that 1 thing, it LGTM

src/main/steps/applicant1/check-your-answers/get.ts
linusnorton: I'm okay with it not being reset in the case data. If that becomes a problem we can think of a better way

velimd (approved):

PR #1005, opened 2021-08-11
NFDIV-963 Applicant2-not-broken event

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-963

### Change description ###
Trigger APPLICANT_2_NOT_BROKEN event when applicant2ScreenHasUnionBroken is No

```
[ ] Yes
[x] No
```


src/main/steps/home/get.ts
```diff
@@ -32,8 +33,10 @@ export class HomeGetController {
       );
     }
 
-    if (req.session.userCase.state === State.Applicant2Approved) {
-      res.redirect(`${CONFIRM_JOINT_APPLICATION}`);
+    if (req.session.userCase.applicant2ScreenHasUnionBroken === YesOrNo.NO) {
```
velimd, line 36: i feel like this should be an event which gets checked. what if the user clicked no but didnt end the case?

lara-mpt: @velimd yeah I see what you're saying, do you think it would be better to have a state change when applicant 2 ends the application on the 'you-cannot-apply' page? 
Or maybe add a field to check if applicant2 has ended the application?

@linusnorton

linusnorton: I guess we could set it to AwaitingApplicant1 and then also check for this field

PR #1012, opened 2021-08-12
better handling when under load

linusnorton:


src/main/server.ts
```diff
@@ -54,6 +55,7 @@ new OidcMiddleware().enableFor(app);
 new StateRedirectMiddleware().enableFor(app);
 new Routes().enableFor(app);
 new ErrorHandler().handleNextErrorsFor(app);
+new TooBusy().enableFor(app);
```
velimd, line 58: should it not be before routes?

velimd (approved):

PR #1024, opened 2021-08-18
NFDIV-1066: Set up PCQ integration

tomxelliott:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1066


package.json
```diff
@@ -97,7 +97,8 @@
     "ts-node": "^10.2.0",
     "tsconfig-paths": "^3.10.1",
     "typescript": "^4.3.5",
-    "uuid": "^8.3.2"
+    "uuid": "^8.3.2",
+    "uuidv4": "^6.2.11"
```
velimd, line 101: if we have this, do we need both?

velimd (approved):

src/main/steps/applicant1/equality/createToken.ts
```diff
@@ -0,0 +1,32 @@
...
+import config from 'config';
+
+const logger = Logger.getLogger('createToken');
+
+const algorithm = 'aes-256-gcm';
+const bufferSize = 16;
+const iv = Buffer.alloc(bufferSize, 0); // Initialization vector.
+const keyLen = 32;
+
+export const createToken = (params: Record<string, unknown>): string => {
```
linusnorton, line 13: Can we do better than `Record<string, unknown>` here? I'm not really sure what params is supposed to be

src/main/steps/applicant1/equality/createToken.ts
```diff
@@ -0,0 +1,32 @@
...
+    });
+    const strParams = JSON.stringify(params);
+    const cipher = crypto.createCipheriv(algorithm, key, iv);
+    encrypted = cipher.update(strParams, 'utf8', 'hex');
+    encrypted += cipher.final('hex');
+  } else {
+    logger.error('PCQ token key is missing.');
+  }
+
+  return encrypted;
```
linusnorton, line 31: I'm not quite sure I like the way this method can silently fail, how does the calling code deal with that?

src/main/steps/applicant1/equality/createToken.ts
```diff
@@ -0,0 +1,32 @@
...
+const algorithm = 'aes-256-gcm';
+const bufferSize = 16;
+const iv = Buffer.alloc(bufferSize, 0); // Initialization vector.
+const keyLen = 32;
+
+export const createToken = (params: Record<string, unknown>): string => {
+  const tokenKey: string = config.get('services.equalityAndDiversity.tokenKey');
+  let encrypted = '';
+
+  if (tokenKey) {
```
linusnorton, line 17: do you need this check? is there ever a situation where the tokenKey is not set?

src/main/steps/applicant1/equality/createToken.ts
tomxelliott: If the token is returned as empty or there are any issues the user will see the PCQ unavailable page

src/main/steps/applicant1/equality/createToken.ts
tomxelliott: Yeah can create a PCQ params type or something like that

src/main/steps/applicant1/equality/createToken.ts
tomxelliott: Only if there is an issue getting it from the config but I am happy to remove this conditional

PR #1035, opened 2021-08-23
Set default app2 name

linusnorton:
Set app2 name when case is linked so that save and close will work properly from the start


src/main/steps/applicant2/enter-your-access-code/post.ts
```diff
@@ -19,6 +19,9 @@ export class AccessCodePostController {
 
     const { saveAndSignOut, saveBeforeSessionTimeout, _csrf, ...formData } = this.form.getParsedBody(req.body);
 
+    formData.respondentUserId = req.session.user.id;
```
velimd, line 22: this is being repeated

PR #1046, opened 2021-08-24
NFDIV-1066

tomxelliott:
Follow host approach used by OIDC middleware to resolve civil partnership returnUrl issue


velimd (approved):
will need to clean this for local but approved for aat

velimd (approved):

PR #1107, opened 2021-09-03
clean up e2e test

velimd:
### Change description ###

```
[ ] Yes
[X] No
```


src/test/steps/IdamUserManager.ts
```diff
@@ -50,11 +50,15 @@ export class IdamUserManager {
     }
   }
 
+  async clearAndKeepOnlyOriginalUser(): Promise<void> {
+    this.users = new Set([...this.users][0]);
```
linusnorton, line 54: new Set(this.users[0])?

velimd: creates a new set with the first element from the previous users.

velimd: i guess i can use the method below getUsername(0) to also get it.

linusnorton: no it's okay, I didn't realise it was a set

linusnorton (approved):

PR #1117, opened 2021-09-07
Fix app2 home get controller and clean up

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-992

### Change description ###
Fix app2 log in when case in Holding state, and clean up the get controller 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/steps/home/get.ts
+++ b/src/main/steps/home/get.ts
@@ -41,53 +41,40 @@ export class HomeGetController {
 }
 
 const applicant1RedirectPageSwitch = (caseState: State, userCase: Partial<Case>, isFirstQuestionComplete: boolean) => {
-  let redirectLink;
-
   switch (caseState) {
     case State.AwaitingApplicant1Response: {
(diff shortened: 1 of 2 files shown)


tomxelliott (approved):

src/main/steps/home/get.ts
```diff
@@ -28,40 +30,64 @@ export class HomeGetController {
...
+    case State.Holding: {
+      redirectLink = HUB_1;
+      break;
+    }
+    default: {
+      redirectLink = isFirstQuestionComplete ? CHECK_ANSWERS_URL : YOUR_DETAILS_URL;
     }
   }
-}
+  return redirectLink;
```
velimd, line 71: might be better to just return in the switch case

velimd: also applies for the other method. this way you dont need a variable set, just return the url in each case and default

velimd (approved):

PR #1147, opened 2021-09-13
NFDIV-770

tomxelliott:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-770


src/main/steps/applicant1/application-ended/get.ts
```diff
@@ -0,0 +1,44 @@
...
+  constructor() {
+    super(__dirname + '/template', generateContent);
+  }
+
+  public async get(req: AppRequest, res: Response): Promise<void> {
+    const caseworkerUser = await getSystemUser();
+    req.locals.api = getCaseApi(caseworkerUser, req.locals.logger);
+
+    try {
+      req.session.userCase.applicationType = ApplicationType.SOLE_APPLICATION;
```
velimd, line 24: i feel like this whole logic should be in the post controller for the previous page.
user clicks switch to sole -> trigger event in post -> redirect to application ended.

tomxelliott: The flow that leads the user to this page is the log in page that is handled by Idam. The user logs in after app2 has said No the marriage hasn't broken down. Therefore there is no post on a page prior to this - only the log in page. If there was was a page before then, yes I would have done that.

src/main/steps/applicant1/application-ended/get.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import * as caseApi from '../../../app/case/CaseApi';
+import { generatePageContent } from '../../common/common.content';
+
+import { generateContent } from './content';
+import ApplicationEndedGetController from './get';
+
+const getSystemUserMock = jest.spyOn(oidc, 'getSystemUser');
+const getCaseApiMock = jest.spyOn(caseApi, 'getCaseApi');
+
+describe('AccessCodeGetController', () => {
```
lara-mpt, line 14:
```suggestion
describe('ApplicationEndedGetController', () => {
```

src/main/steps/applicant1/application-ended/get.ts
velimd: fair

velimd (approved):

lara-mpt (approved):

PR #1159, opened 2021-09-15
Nfdiv 1017

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1017

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/modules/state-redirect/index.ts
+++ b/src/main/modules/state-redirect/index.ts
@@ -1,9 +1,20 @@
+import dayjs from 'dayjs';
+import customParseFormat from 'dayjs/plugin/customParseFormat';
 import { Application, NextFunction, Response } from 'express';
 
+import { CaseWithId } from '../../app/case/case';
 import { State } from '../../app/case/definition';
 import { AppRequest } from '../../app/controller/AppRequest';
 import { PaymentModel } from '../../app/payment/PaymentModel';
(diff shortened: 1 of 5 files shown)


tomxelliott (approved):

src/main/modules/state-redirect/index.ts
```diff
@@ -37,4 +56,12 @@ export class StateRedirectMiddleware {
       })
     );
   }
+
+  private hasPartnerResponded(userCase: CaseWithId, isApplicant2: boolean) {
+    return !(
+      ((isApplicant2 && [State.AwaitingApplicant1Response, State.Applicant2Approved].includes(userCase?.state)) ||
+        (!isApplicant2 && userCase?.state === State.AwaitingApplicant2Response)) &&
+      dayjs(userCase.dueDate, 'MMMM Do YYYY').diff(dayjs()) < 0
+    );
```
lara-mpt, line 65: would it be easier to make this be hasPartnerNotResponded rather than negate it? 
 because if it returns True at the moment it doesn't necessarily mean they've responded, just that they haven't not responded

velimd: i agree to the name change, the ! was done before i writed the logic/check and i negated it where the method is called [here](https://github.com/hmcts/nfdiv-frontend/pull/1159/files#diff-631965ac66b32e6255fecaff8f97fda20f61ce22335af920dd9f2bd642c9942aR31) but cleaning the name up is better.Thanks

lara-mpt (approved):

PR #1199, opened 2021-09-23
Updated definition ts with latest changes related to keep contact details confidential field

nitinprabhu:
### JIRA link ###

https://tools.hmcts.net/jira/browse/NFDIV-1192

### Change description ###

-  contact details confidential field type is being changed from drop down to yes or no field see ref PR https://github.com/hmcts/nfdiv-case-api/pull/733


src/main/app/case/to-api-format.ts
```diff
@@ -149,7 +149,7 @@ const addressPrivateYesNoOrNull = (value: YesOrNo) => {
   if (!value) {
     return null;
   }
-  return value === YesOrNo.YES ? ConfidentialAddress.KEEP : ConfidentialAddress.SHARE;
+  return value;
```
linusnorton, line 152: I think you could get rid of this method now and just use the standard YesOrNo conversion

nitinprabhu: done https://github.com/hmcts/nfdiv-frontend/pull/1199/commits/fe198e567c09e50c9a7f84aca544f4ef765a81ea

PR #1204, opened 2021-09-24
NFDIV-770

tomxelliott:
Remove applicant 2 address if they have said it is private when switching to sole


src/main/app/case/answers/possibleAnswers.ts
```diff
@@ -47,3 +47,18 @@ export const getUnreachableAnswersAsNull = (userCase: Partial<Case>): Partial<Ca
     everyField.filter(key => !possibleAnswers.includes(key) && userCase[key]).map(key => [key, null])
   );
 };
+
+export const deleteAddressIfPrivate = (userCase: Partial<Case>): void => {
+  if (
+    userCase.applicationType === ApplicationType.SOLE_APPLICATION &&
+    userCase.applicant2AddressPrivate === YesOrNo.YES
+  ) {
+    delete userCase.applicant2Address1;
```
velimd, line 56: wouldnt these get removed in the unreachable answers?

velimd (commented):
other than my comment, it LGTM

src/main/app/case/answers/possibleAnswers.ts
tomxelliott: Nope, they are part of the applicant 1 Sole journey

velimd: but when they select private for the address, they dont see this page as the flow shouldnt go there right? (the step sequence should skip the page meaning these fields would get set to null, na?)

velimd (approved):

src/main/app/case/answers/possibleAnswers.ts
lara-mpt: I don't think the fields are currently unset by the unreachableAnswers because of 
[https://github.com/hmcts/nfdiv-frontend/blob/8320b95a29ad1e33d11ad19d27125308abafb90b/src/main/steps/applicant1Sequence.ts#L343](url)

applicant1KnowsApplicant2Address will be null so ENTER_THEIR_ADDRESS will be reached by the applicant 1 sole journey

if we changed that to 
`data.applicant1KnowsApplicant2Address === YesOrNo.YES ? ENTER_THEIR_ADDRESS: NEED_TO_GET_ADDRESS,`
then they should be unset

PR #1220, opened 2021-09-28
Nfdiv-1212 Fix app2 HomeGetController

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1212

### Change description ###
fix app2 HomeGetController - redirect app2 users to CHECK_JOINT_APPLICATION if all questions completed as per the app 2 flow

```
[ ] Yes
[x] No
```


velimd (commented):
im not sure about this implementation. Its okay but i feel like it would be better if we checked if the next incomplete answer (forgot the name of the function) was completed then we can direct them to check joint application.

velimd (approved):

PR #1306, opened 2021-10-13
clean up other court cases

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1052

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
     "uglifyjs-webpack-plugin": "^2.2.0",
     "webdriverio": "^7.14.1",
     "webpack": "^5.58.2",
-    "webpack-cli": "^4.9.0",
+    "webpack-cli": "^4.9.1",
     "webpack-dev-middleware": "^5.2.1",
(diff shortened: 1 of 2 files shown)


src/main/app/case/answers/getAnswerRows.test.ts
```diff
@@ -140,9 +140,7 @@ describe('getAnswerRows()', () => {
         applicant1FullNameOnCertificate: 'Sarah Smith',
         applicant2FullNameOnCertificate: 'Billy Bob',
         applicant1LegalProceedings: YesOrNo.YES,
-        applicant1LegalProceedingsRelated: ['marriage', 'property'],
         applicant2LegalProceedings: YesOrNo.YES,
```
lara-mpt, line 143: could you get rid of applicant1LegalProceedings and applicant2LegalProceedings as well?

lara-mpt: oh right, I didn't think they were

PR #1351, opened 2021-10-21
removed required value

danotade:
### JIRA link ###
Updating the content on financial order page

### Change description ###
Content refresh needed. No options shown when 'Yes' is selected for BOTH for sole or joint

```
[ ] Yes
[ x] No
```


linusnorton (approved):

velimd (commented):
if the field is no long in use, then it should be completely removed.

PR #1371, opened 2021-10-25
Nfdiv 1259 amend contact details page

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1295

### Change description ###
Amend contact details page

```
[ ] Yes
[x] No
```


src/main/steps/applicant1/check-contact-details/content.ts
```diff
@@ -0,0 +1,35 @@
...
+const cy = en;
+
+const languages = {
+  en,
+  cy,
+};
+
+export const generateContent: TranslationFn = content => {
+  const address = [content.formState?.applicant1Address1];
+  address.push(content.formState?.applicant1Address2);
```
linusnorton, line 19: why is formState using ?. isn't it always set?

src/main/steps/applicant1/check-contact-details/content.ts
```diff
@@ -0,0 +1,35 @@
...
+  address.push(content.formState?.applicant1Address3);
+  address.push(content.formState?.applicant1AddressTown);
+  address.push(content.formState?.applicant1AddressCounty);
+  address.push(content.formState?.applicant1AddressPostcode);
+  address.push(content.formState?.applicant1AddressCountry);
+  const applicantAddress = address.filter(Boolean).join('<br>');
```
linusnorton, line 25: I had no idea that works

src/main/steps/applicant1/check-contact-details/template.njk
```diff
@@ -0,0 +1,63 @@
...
+  {% endblock %}
+
+  {{ govukSummaryList({
+    rows: [
+      {
+        key: {
+          text: yourAddress
+        },
+        value: {
+          text: applicantAddress | safe
```
linusnorton, line 20: how come that needs a | safe?

src/main/steps/applicant2/check-contact-details/content.ts
```diff
@@ -0,0 +1,37 @@
...
+const cy = en;
+
+const languages = {
+  en,
+  cy,
+};
+
+export const generateContent: TranslationFn = content => {
+  const address = [content.formState?.applicant2Address1];
+  address.push(content.formState?.applicant2Address2);
```
linusnorton, line 20: see previous comment

src/main/steps/applicant1/check-contact-details/content.ts
velimd: i believe its the way we do generateContent, as we are passing in a partial<caseData> its complaining it might be empty.

linusnorton: if the formState is partial then then ?. should be on applicant1Address2 but you're doing an undefined check on formState

src/main/steps/applicant1/check-contact-details/content.ts
velimd: yh, ill clean it up, i believe we do something similar in check your answers.

src/main/steps/applicant1/check-contact-details/template.njk
velimd: because the address is using br.

src/main/steps/applicant2/check-contact-details/content.ts
velimd: same as before, we are passing partial case so its complaining it might be undefined.

src/main/steps/applicant1/check-contact-details/content.ts
velimd: oh, looking into it, its also a case where formstate is optional in commonContent function, looks like for the error controller as it dont need it there to be passed, we can turn that off and pass it in.

linusnorton: I would always set it even if it's not needed. Saves using ?. everywhere

velimd: kl, ill do that tho i think that should be another pr, changing it here will make this pr even bigger.

linusnorton: sure, do it in another PR

linusnorton (approved):

src/main/steps/applicant1/check-phone-number/content.ts
```diff
@@ -0,0 +1,47 @@
+import { TranslationFn } from '../../../app/controller/GetController';
+import { FormContent } from '../../../app/form/Form';
+import { isPhoneNoValid } from '../../../app/form/validation';
+
+const en = () => ({
+  title: 'Your phone number',
+  line1:
+    'Enter your phone number so court staff can contact you quickly, if they need to. ' +
+    'You can choose to keep your phone number private from your later in this application.',
```
benouaer, line 9: is there something missing in that second sentence?

src/main/steps/applicant1/hub-page/right-column/content.ts
```diff
@@ -15,7 +15,8 @@ const en = ({ isDivorce, isApplicant2 }: CommonContent) => ({
     isDivorce ? 'divorce application' : 'application to end your civil partnership'
   } (PDF)</a>`,
   reviewContactDetails: `<a class="govuk-link" href="${
-    (isApplicant2 ? APPLICANT_2 : '') + REVIEW_CONTACT_DETAILS
+    (isApplicant2 ? (userCase?.applicationType === ApplicationType.SOLE_APPLICATION ? RESPONDENT : APPLICANT_2) : '') +
```
benouaer, line 18: is there a valid scenario where applicationType can be null, if not we can remove '?'

src/test/functional/features/applicant2.feature
```diff
@@ -91,6 +91,8 @@ Feature: Applicant 2
     And I click "I believe that the facts stated in this application are true"
     When I click "Continue"
     Then the page URL should be "/pay-and-submit"
+    Given I pay and submit the joint application
```
benouaer, line 94: this should probably be 'When' rather than 'Given'

src/test/functional/features/update-contact-details.feature
```diff
@@ -0,0 +1,74 @@
...
+    And I've already completed the form using the fixture "completeCase"
+
+  Scenario: Applicant 1 updates contact details
+    And I go to '/check-your-answers'
+    And I click "I confirm"
+    And I click "I believe that the facts stated in this application are true"
+    When I click "Continue to payment"
+    And I pay and submit the application
+    Then the page should include "Application submitted"
+    Given a case worker issues the application
```
benouaer, line 13: can we leave a line before each 'Given', might make it easier to follow

src/test/functional/features/update-contact-details.feature
```diff
@@ -0,0 +1,74 @@
...
+    And I select "Building and street"
+    And I type "Address 1"
+    And I select "Town or city"
+    And I type "LONDON"
+    And I select "County"
+    And I type "CITY OF WESTMINSTER"
+    And I select "Postcode"
+    And I type "SW1H 9AA"
+    And I select "Country"
+    And I type "UK"
```
benouaer, line 31: could we do the whole address as one step

src/main/steps/applicant1/hub-page/right-column/content.ts
velimd: working on the ? in another branch

src/test/functional/features/update-contact-details.feature
velimd: Maybe its easier if we use When, i dont like using a gap, doesnt feel like its part of the test but another section.

src/test/functional/features/update-contact-details.feature
velimd: we can create a function but really, this is how the user will do stuff.

PR #1391, opened 2021-10-28
remove stub case in post controller

velimd:
### Change description ###
Not sure why this is here, probably for test and those who use the example as the postcode but still it shouldnt be here.

```
[ ] Yes
[X] No
```


danotade (approved):

linusnorton:
This was originally added by Seb because the postcode API said not to use it for functional tests but I'm happy to remove it

velimd:
> This was originally added by Seb because the postcode API said not to use it for functional tests but I'm happy to remove it

I assume so as well, i might change the name to testStubApi or add a note.

PR #1419, opened 2021-11-09
Nfdiv 1296 Remove mutable form state

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1296

### Change description ###
Remove mutable form state

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/form/Form.ts
+++ b/src/main/app/form/Form.ts
@@ -125,7 +125,7 @@ export type ValidationCheck = (
   formData: Partial<Case>
 ) => void | string;
 export type FormFields = Partial<Record<string, FormField>>;
-export type FormFieldsFn = (formState: Partial<Case>) => FormFields;
+export type FormFieldsFn = (userCase: Partial<Case>) => FormFields;
 
 export interface FormContent {
   submit: {
--- a/src/main/app/form/Form.ts
+++ b/src/main/app/form/Form.ts
@@ -4,26 +4,13 @@ import { AnyObject } from '../controller/PostController';
 import { setupCheckboxParser } from './parser';
 
 export class Form {
-  constructor(private readonly form: FormContent, private formState: Partial<Case> = {}) {}
-
-  public setFormState(formState: Partial<Case>): void {
-    this.formState = formState;
-  }
-
-  public getFields(checkFields?: FormContent['fields']): FormContent['fields'] {
-    const fields = checkFields || this.form?.fields;
-    if (typeof fields === 'function') {
-      return fields(this.formState);
-    }
-
-    return fields;
-  }
+  constructor(private readonly fields: FormFields | FormFieldsFn) {}
 
   /**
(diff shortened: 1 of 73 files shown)


src/main/app/form/Form.ts
```diff
@@ -4,26 +4,13 @@ import { AnyObject } from '../controller/PostController';
...
-    return fields;
-  }
+  constructor(private readonly fields: FormFields | FormFieldsFn) {}
 
   /**
    * Pass the form body to any fields with a parser and return mutated body;
    */
   public getParsedBody(body: AnyObject, checkFields?: FormContent['fields']): Partial<CaseWithFormData> {
-    const fields = this.getFields(checkFields);
+    const fields = checkFields || this.fields;
```
linusnorton, line 13: is checkFields still needed? does anything use it?

src/main/app/form/Form.ts
```diff
@@ -4,26 +4,13 @@ import { AnyObject } from '../controller/PostController';
...
-
-  public getFields(checkFields?: FormContent['fields']): FormContent['fields'] {
-    const fields = checkFields || this.form?.fields;
-    if (typeof fields === 'function') {
-      return fields(this.formState);
-    }
-
-    return fields;
-  }
+  constructor(private readonly fields: FormFields | FormFieldsFn) {}
```
linusnorton, line 7: shouldn't this just be `FormFields` or can a function still be passed?

src/main/modules/nunjucks/index.ts
```diff
@@ -22,9 +22,9 @@ export class Nunjucks {
     });
 
     env.addGlobal('getError', function (fieldName: string): { text?: string } | boolean {
-      const { form, sessionErrors, errors, formState } = this.ctx;
+      const { form, sessionErrors, errors } = this.ctx;
 
-      const hasMoreThanTwoFields = new Form(form, formState).getFieldNames().size >= 2;
+      const hasMoreThanTwoFields = new Form(form.fields).getFieldNames().size >= 2;
```
linusnorton, line 27: I had no idea a form gets constructed here, that's pretty horrible but we'll have to look at it later

src/main/app/form/Form.ts
lara-mpt: it's needed for the subfields

lara-mpt: .map(subField => this.getParsedBody(body, subField))

src/main/app/form/Form.ts
lara-mpt: @linusnorton I don't think a function can still be passed, but if we get rid of the FormFieldsFn here then we would have to add a <FormFields> type assertion whenever we create a new form, and I wasn't sure if that's better?

-     const form = new Form(<FormFields>this.fields);

linusnorton (approved):

PR #1436, opened 2021-11-12
Nfdiv 1286 updating confirm application page

danotade:
### JIRA link ###
https://tools.hmcts.net/jira/secure/RapidBoard.jspa?rapidView=1609&view=planning&selectedIssue=NFDIV-1286&quickFilter=9976&issueLimit=100

### Change description ###
Confirm your joint application page needs to be updated as per the Google docs for both applicants.

```
[ ] Yes
[x ] No
```


src/main/steps/applicant1/connection-summary/content.ts
```diff
@@ -5,28 +5,47 @@ import { CommonContent } from '../../common/common.content';
 
 export const jurisdictionMoreDetailsContent = (
   connections: JurisdictionConnections[] | undefined,
+  isDivorce: boolean,
   isRespondent = false
 ): { connectedToEnglandWales: string; readMore: string } => {
   const resConnection = enContainsHabitualResConnection(connections);
```
lara-mpt, line 11: I think it's worth renaming resConnection now, otherwise it's confusing between that and residualConnection

src/main/steps/applicant1/connection-summary/content.ts
```diff
@@ -58,6 +77,20 @@ const enDomicile = {
...
+const getResidual = (isDivorce: boolean) => {
+  return {
+    helpText12: 'Residual',
+    helpText13:
+      'If you’re in a same-sex couple and if none of the other connections apply, the court may still have jurisdiction if: ',
+    helpText14: `<ul class="govuk-list govuk-list--bullet"><li class="govuk-list govuk-list--bullet">you ${
+      isDivorce ? 'married' : 'formed your civil partnership'
+    } in England or Wales and</li>
+      <li class="govuk-list govuk-list--bullet">it would be in the interests of justice for the court to consider the application. For example, your home country does not allow ${
+        isDivorce ? 'divorce' : 'civil partnerships to be ended] between same-sex couples'
```
lara-mpt, line 89:
```suggestion
        isDivorce ? 'divorce' : 'civil partnerships to be ended'
```

src/main/steps/applicant1/connection-summary/content.ts
```diff
@@ -106,6 +145,7 @@ const en = ({ isDivorce, partner, userCase }: CommonContent) => {
     readMore: 'Read more about your connections',
     ...enContainsHabitualResConnection(userCase?.connections),
     ...enContainsDomConnection(userCase?.connections),
+    ...getResidual(isDivorce),
```
lara-mpt, line 148: shouldn't this be '...enContainsResidualConnection(userCase?.connections, isDivorce)' ?

src/main/steps/applicant1/confirm-your-joint-application/content.ts
```diff
@@ -1,158 +1,180 @@
...
+        }:` +
+        '<br>' +
+        userCase?.applicant1LegalProceedingsDetails +
+        '<br><br>' +
+        `Applicant 2 has given details of other court cases relating to the ${
+          isDivorce ? 'marriage' : 'civil partnership'
+        }:` +
+        '<br>' +
+        userCase?.applicant2LegalProceedingsDetails
+      : !bothApplicantsLegalProceedings(userCase)
```
lara-mpt, line 84: !bothApplicantsLegalProceedings will return True if **either** applicant1LegalProceedings or applicant2LegalProceedings is No, so it won't work for the next line which is for if **both** are No

src/main/steps/applicant1/connection-summary/content.ts
danotade: Agree.  Changed

src/main/steps/applicant1/connection-summary/content.ts
danotade: Oops!

src/main/steps/applicant1/confirm-your-joint-application/content.ts
```diff
@@ -1,158 +1,176 @@
...
-import { connectionBulletPointsTextForSoleAndJoint } from '../../../app/jurisdiction/bulletedPointsContent';
+import { connectionBulletPointsTextForRespondent } from '../../../app/jurisdiction/bulletedPointsContent';
 import { jurisdictionMoreDetailsContent } from '../../../steps/applicant1/connection-summary/content';
-import { Sections } from '../../applicant1Sequence';
-import { moreDetailsComponent } from '../../applicant2/check-your-joint-application/content';
 import { CommonContent } from '../../common/common.content';
-import * as urls from '../../urls';
 
-const en = ({ isDivorce, partner, userCase, marriage, civilPartnership }: CommonContent) => ({
+const isSubmit = (isApplicant2: boolean, userCase: Partial<CaseWithId> | undefined): boolean => {
```
lara-mpt, line 13: how comes userCase can be undefined here?

src/main/steps/applicant1/confirm-your-joint-application/content.ts
```diff
@@ -1,158 +1,176 @@
...
+  return (
+    isApplicant2 ||
+    (!isApplicant2 &&
```
lara-mpt, line 16: since you checked if it's **isApplicant2** above then I don't think you don't need to add in the **!isApplicant2** bit as well

src/main/steps/applicant1/confirm-your-joint-application/content.ts
danotade: Actually, can userCase ever be undefined?  This has been confusing me for a while.  What situation can a userCase be undefined?

src/main/steps/applicant1/confirm-your-joint-application/content.ts
```diff
@@ -1,158 +1,176 @@
...
+        }:` +
+        '<br>' +
+        userCase?.applicant1LegalProceedingsDetails +
+        '<br><br>' +
+        `Applicant 2 has given details of other court cases relating to the ${
+          isDivorce ? 'marriage' : 'civil partnership'
+        }:` +
+        '<br>' +
+        userCase?.applicant2LegalProceedingsDetails
+      : userCase?.applicant1LegalProceedings === YesOrNo.NO && userCase?.applicant2LegalProceedings === YesOrNo.NO
```
lara-mpt, line 80: instead of adding a separate 'No and No' check you could run through 'Yes and Yes' and then 'Yes' either way and then finally the else part would be the 'No and No'

if that makes any sense

danotade: That's more of a preference thing. I think it works both ways in this case.  I find it easier to understand in my way of thinking

src/main/steps/applicant1/confirm-your-joint-application/content.ts
lara-mpt: no, as far as I'm aware it won't ever be undefined

danotade: So why do we have the ? everywhere next to it?   OK so I will remove the undefined bit but it's even more confusing

src/main/steps/applicant1/confirm-your-joint-application/content.ts
danotade: Again preference.  Yes you are right but I do find it easier to read that I'm not dealing with Applicant 2 by putting that there (so I'm dealing with Applicant 1).  Easier to read in my head.  I have removed though

src/main/steps/applicant1/confirm-your-joint-application/content.ts
lara-mpt: I think @velimd is addressing that

danotade: I've changed it but it looks so weird now.  In a couple of places it looks like it cannot be undefined (userCase!)  and in others it looks like it can userCase?  Perhaps we should leave it as it it with possibly undefined, then create a ticket that removes it.  What do you think?

lara-mpt: I think Veli's working on getting rid of the ? so I would assume that would also fix this problem ^

danotade: OK so is it better to leave it as it is showing it can be undefined - even though we know it never will be - or add the ! in certain places?  I prefer to leave as it is and let Veli remove all the undefined as opposed to mixing ? and ! in the code next to userCase

src/main/steps/applicant1/confirm-your-joint-application/template.njk
```diff
@@ -1,37 +1,81 @@
-{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+{% from "govuk/components/details/macro.njk" import govukDetails %}
```
velimd, line 1: this whole template seems to be similar to review the application template, maybe you should reuse that.

src/main/steps/applicant1/connection-summary/content.ts
```diff
@@ -5,28 +5,47 @@ import { CommonContent } from '../../common/common.content';
...
-  const resConnection = enContainsHabitualResConnection(connections);
+  const habConnection = enContainsHabitualResConnection(connections);
   const domConnection = enContainsDomConnection(connections);
+  const residualConnection = enContainsResidualConnection(connections, isDivorce);
 
-  const connectionIndex = isRespondent || (resConnection && domConnection) ? 2 : domConnection ? 1 : 0;
+  const connectionIndex =
+    isRespondent || (habConnection && domConnection && residualConnection)
+      ? 3
+      : domConnection
```
velimd, line 18: this logic doesnt seem to consider what it was before, did something change? before hab and dom were true, its 2 but here if its true itll automatically go to 1 (considering residual is false).

src/main/steps/applicant2/confirm-your-joint-application/content.ts
```diff
@@ -21,7 +21,6 @@ export const form: FormContent = {
   fields: {
     applicant2IConfirmPrayer: {
       type: 'checkboxes',
-      label: l => l.confirm,
       labelSize: 'm',
```
velimd, line 24: how comes the label is not here. itll just be an empty checkbox no?

src/main/steps/applicant1/confirm-your-joint-application/content.ts
velimd: its considered undefined currently because its an optional property (which im making it conpulsory), hence all the ?.

src/main/steps/applicant1/confirm-your-joint-application/template.njk
danotade: I changed.  I copied the review application template and changed to fit what I need for this one

PR #1470, opened 2021-11-19
Nfdiv 1346

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1346

### Change description ###

```
[ ] Yes
[X] No
```


src/main/steps/applicant1/hub-page/sole/content.ts
```diff
@@ -110,14 +110,15 @@ export const generateContent: TranslationFn = content => {
...
     State.AosOverdue,
     State.Holding,
-    State.PendingDispute,
-    State.Disputed,
+    State.AwaitingGeneralConsideration,
     State.AwaitingLegalAdvisorReferral,
     State.AwaitingPronouncement,
     State.FinalOrderComplete,
   ].indexOf(content.userCase.state as State);
+  const applicationDisputing = content.userCase.confirmDisputeApplication === YesOrNo.YES;
```
lara-mpt, line 118: I think it would be better to base the check off of the disputeApplication field, just to keep it consistent with other checks which use that field

lara-mpt (approved):

src/main/steps/applicant1/hub-page/sole/content.ts
velimd: i saw some other place use it as well so i changed it there as well.

[commit](https://github.com/hmcts/nfdiv-frontend/pull/1470/commits/09410b41180dd895ad93b60bcdd1f8be0edab2d5)

PR #1476, opened 2021-11-22
Fix jurisdiction fields

lara-mpt:
### Change description ###
Set jurisdiction fields as null when they are unreachable and clean up

```
[ ] Yes
[x] No
```


velimd (commented):
is there a specific reason these get set to null? if so, maybe we can do it like the others were we set it to null in to-api-format.

lara-mpt:
> is there a specific reason these get set to null? if so, maybe we can do it like the others were we set it to null in to-api-format.

I tried setting it in the to-api-format but the different criteria for how/when these fields can be set gets quite complicated, so it was much easier to use the getAllPossibleAnswersForPath to figure out which fields can be set

src/main/app/jurisdiction/JurisdictionPostController.ts
```diff
@@ -20,4 +23,30 @@ export class JurisdictionPostController extends PostController<AnyObject> {
...
+    'applicant1DomicileInEnglandWales',
+    'applicant2DomicileInEnglandWales',
+    'bothLastHabituallyResident',
+    'applicant1LivingInEnglandWalesTwelveMonths',
+    'applicant1LivingInEnglandWalesSixMonths',
+    'jurisdictionResidualEligible',
+  ];
+
+  const possibleAnswers = getAllPossibleAnswersForPath(userCase, stepsWithContentApplicant1);
+  jurisdictionFields.forEach(field => (!possibleAnswers.includes(field) ? (userCase[field] = null) : userCase[field]));
```
linusnorton, line 49: that else statement is not really doing anything and it kind of feels like dancing around using a for statement. I would probably be inclined to go for 

```
for (const field of jurisdictionFields) {
  if (!possibleAnswers.includes(field)) {
    userCase[field] = null;
  }
}
```

but equally you could use reduce if you want to do some fancy functional stuff.

linusnorton (approved):

PR #1490, opened 2021-11-24
Fix residual jurisdiction field

lara-mpt:
### Change description ###
Fix residual jurisdiction field

```
[ ] Yes
[x] No
```


tomxelliott (approved):
Tested against AAT and all works as expected!!

PR #1495, opened 2021-11-25
Nfdiv 1415 2nd implementation

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1415

### Change description ###
This is for when the user is already in a linked joint application but when the case is AwaitingApplicant1Response and app1 decides to switch to sole and submit.

```
[ ] Yes
[X] No
```


src/main/steps/applicant1/enter-their-address/get.ts
```diff
@@ -0,0 +1,27 @@
...
+import { GetController } from '../../../app/controller/GetController';
+
+@autobind
+export default class EnterTheirAddressGetController extends GetController {
+  public async get(req: AppRequest, res: Response): Promise<void> {
+    if (
+      req.session.userCase.applicant2AddressPrivate === YesOrNo.YES &&
+      req.session.userCase.state === State.AwaitingApplicant1Response
+    ) {
+      req.session.userCase.applicant2Address1 = undefined;
```
benouaer, line 15: is this to hide applicant2 details from applicant1? shouldn't we just obfuscate that in the template rather than actually changing the data

velimd: the template uses the session data so im not sure how we should do that.

velimd: maybe a custom function which gets the userCase and we override it in the custom controller?

PR #1507, opened 2021-11-30
Nfdiv-1474 fix and update states for sole hub page

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1474

### Change description ###
Fix and update states for sole hub page

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/steps/applicant1/hub-page/progress-bar.njk
+++ b/src/main/steps/applicant1/hub-page/progress-bar.njk
@@ -43,32 +43,32 @@
     label: {
       text: response
     },
-    complete: progressionIndex > 4,
-    active: progressionIndex >= 0 and progressionIndex < 5
+    complete: progressionIndex > 7,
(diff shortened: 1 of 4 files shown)


src/main/steps/applicant1/hub-page/progress-bar.njk
```diff
@@ -43,32 +43,32 @@
     label: {
       text: response
     },
-    complete: progressionIndex > 4,
-    active: progressionIndex >= 0 and progressionIndex < 5
+    complete: progressionIndex > 7,
```
velimd, line 46: you added 5 different states but this went up by 3, is this correct?

lara-mpt: @velimd yeah that was the main fix for the ticket, the response circle wasn't showing up as complete when the state was Holding because the number was 2 too high

velimd (approved):

PR #1511, opened 2021-12-01
Nfdiv 1311

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1311

### Change description ###

```
[ ] Yes
[X] No
```


src/main/modules/document-download/index.ts
```diff
@@ -28,18 +19,36 @@ export class DocumentDownloadMiddleware {
...
+      },
+    };
+
+    const dmStoreProxyForCertificateOfServicePdf = {
+      endpoints: ['/downloads/certificate-of-service'],
+      path: (req: AppRequest) => {
+        return req.session.userCase.alternativeServiceOutcomes.find(
+          doc =>
+            doc.value.successfulServedByBailiff === YesOrNo.YES &&
+            doc.value.certificateOfServiceDocument.documentType === DocumentType.CERTIFICATE_OF_SERVICE
```
lara-mpt, line 43: do you know that the CERTIFICATE_OF_SERVICE is the correct document? Because my ticket also stated it was the CERTIFICATE_OF_SERVICE but it was DEEMED_AS_SERVICE_GRANTED and DISPENSE_WITH_SERVICE_GRANTED

also since the deemed and dispensed are both also certificates of service it might be good to specify that that download link is specifically for bailiff certificate of service:
e.g '/bailiff-certificate-of-service'

src/main/steps/applicant1/hub-page/right-column/content.ts
```diff
@@ -47,8 +49,12 @@ export const generateContent: TranslationFn = content => {
   const statesWithoutRespondentAnswers = [State.AwaitingAos, State.AosDrafted, State.AosOverdue];
   const aosSubmitted =
     !content.isJointApplication && !statesWithoutRespondentAnswers.includes(<State>content.userCase.state);
+  const hasCertificateOfService = content.userCase.alternativeServiceOutcomes?.find(
+    alternativeServiceOutcome => alternativeServiceOutcome.value.successfulServedByBailiff === YesOrNo.YES
```
lara-mpt, line 53: there's quite a lot of overlap with our two prs, as I used 
  const hasCertificateOfService =
    content.userCase.alternativeServiceOutcomes?.length &&
    content.userCase.alternativeServiceOutcomes?.[0].value.alternativeServiceType !== AlternativeServiceType.BAILIFF;
 
just to note that one of us will have to sort out the conflicts for it

lara-mpt: also your .find() works better so I'll update my pr to use that

src/main/modules/document-download/index.ts
velimd: how do you confirm this?
i guess i can use the same find in my content and check if bailiff was successful. not sure about anything else as when i went through this service, it was CERTIFICATE_OF_SERVICE.

src/main/steps/applicant1/hub-page/right-column/content.ts
velimd: thats fine, we can clean it up based on which ever branch is in first. i can clean up if you want. main issue, why is CERTIFICATE_OF_SERVICE common, is it a different document with the same name or the same document, if its the same, then we just gotta show it for any condition tbh as long as the document exist?

lara-mpt (approved):

PR #1515, opened 2021-12-03
Nfdiv 1289

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1289

### Change description ###

```
[ ] Yes
[X] No
```


src/main/steps/applicant1/hub-page/sole/form.njk
```diff
@@ -0,0 +1,15 @@
...
+    {% block form_content %}
+      {% include "common/form/fields.njk" %}
+    {% endblock %}
+
+    {{ govukButton({
+      text: conditionalOrder.applyForConditionalOrder,
+      classes: form.submit.classes,
+      type: "submit",
+      preventDoubleClick: true
+    }) }}
```
lara-mpt, line 14: You could set the applicant1ApplyForConditionalOrderStarted field here

(It's one of the fields I'm using for the joint conditional order hub page)

velimd: i couldnt find it in the backend so i removed it. thanks ill add it

src/main/steps/applicant1/hub-page/sole/hub.njk
```diff
@@ -1,11 +1,30 @@
...
 
-{% if progressionIndex == 9 %}
+{% set readMoreContent %}
+  <p class="govuk-body">{{ readMoreSummary }}</p>
+  <ol class="govuk-list govuk-list--number">
+    {% for property, line in readMoreSteps %}
+      <li>{{ line | safe }}</li>
+    {% endfor %}
+  </ol>
+  <p class="govuk-body">{{ holding.line4 | safe }}</p>
```
lara-mpt, line 10: if that `holding.line4` is used multiple times in the template outside of the Holding state content it might be worth just making that a separate variable instead of leaving it inside the Holding content

src/main/steps/home/get.ts
```diff
@@ -78,6 +79,9 @@ const applicant1RedirectPageSwitch = (caseState: State, userCase: Partial<Case>,
     case State.AwaitingGeneralConsideration: {
       return HUB_PAGE;
     }
+    case State.ConditionalOrderDrafted: {
+      return READ_THE_RESPONSE;
+    }
```
lara-mpt, line 84: This is only in the case for a sole application, but you could leave this in and I can update my pr after:

https://github.com/hmcts/nfdiv-frontend/pull/1531/files#diff-461aadf81468465bfe6c5fcfc5e41e1b079e65e9ce1f5d5cae039eedb291514eR79-R81

lara-mpt (approved):

src/main/steps/applicant1/hub-page/post.ts
```diff
@@ -1,11 +1,15 @@
...
 import { AnyObject, PostController } from '../../../app/controller/PostController';
 
 @autobind
 export default class HubPagePostController extends PostController<AnyObject> {
-  protected getEventName(): string {
-    return APPLICANT_1_CONFIRM_RECEIPT;
+  protected getEventName(req: AppRequest<AnyObject>): string {
+    return req.session.userCase.applicationType === ApplicationType.JOINT_APPLICATION &&
+      req.body.applicant1ConfirmReceipt
+      ? APPLICANT_1_CONFIRM_RECEIPT
```
linusnorton, line 12: this doesn't look correct to me... but I might be wrong.

If you submit the from the hub page surely there will be different events based on the state of the case? Drafting the CO, drafting the FO and also confirming receipt. Doing it based on whether it's a joint app doesn't seem right.

src/main/steps/home/get.ts
```diff
@@ -78,6 +79,9 @@ const applicant1RedirectPageSwitch = (caseState: State, userCase: Partial<Case>,
     case State.AwaitingGeneralConsideration: {
       return HUB_PAGE;
     }
+    case State.ConditionalOrderDrafted: {
+      return READ_THE_RESPONSE;
```
linusnorton, line 83: would you know run through the steps to find out what part of the CO they're on?

velimd: i believe at this point itll end on the hub page considering the hub page and the read the response page are just continue buttons.

src/main/steps/applicant1/hub-page/post.ts
velimd: currently theres only 2 events and the logic was based on if its sole or joint. ofc down the line if theres more then itll change but for now its based on that, i can change it to be based on states and application types.

linusnorton (approved):

PR #1531, opened 2021-12-09
Nfdiv 995 joint hub page awaiting conditional order

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-995

### Change description ###
Joint hub page awaiting conditional order

```
[ ] Yes
[x] No
```


src/main/steps/applicant1/hub-page/post.ts
```diff
@@ -1,11 +1,24 @@
...
 export default class HubPagePostController extends PostController<AnyObject> {
-  protected getEventName(): string {
-    return APPLICANT_1_CONFIRM_RECEIPT;
+  protected getEventName(req: AppRequest): string {
+    if (req.session.userCase.state === State.Holding) {
+      return req.session.isApplicant2 ? APPLICANT_2_CONFIRM_RECEIPT : APPLICANT_1_CONFIRM_RECEIPT;
+    } else if (req.session.userCase.state === State.AwaitingConditionalOrder) {
+      return DRAFT_CONDITIONAL_ORDER;
+    } else {
+      return UPDATE_CONDITIONAL_ORDER;
```
linusnorton, line 21: would this ever be called from the hub page?

lara-mpt: if app1 has already started CO they would have triggered DRAFT_CONDITIONAL_ORDER, so the case state would be ConditionalOrderDrafted, when app2 then logs in and starts CO from the hub page they should trigger UPDATE_CONDITIONAL_ORDER as the CO has already been drafted

velimd (approved):

velimd (approved):

src/main/steps/home/get.ts
```diff
@@ -33,18 +34,24 @@ export class HomeGetController {
       throw new Error('Invalid case type');
     }
 
-    const firstQuestionForm = getApplicantFirstQuestionForm(
-      req.session.isApplicant2,
-      req.session.userCase.applicationType!
-    );
+    const firstQuestionFormFields = req.session.isApplicant2
```
velimd, line 37: i think it was better in the function. is there a reason it was moved out?

lara-mpt: The reason I moved it into getApplicant2FirstQuestionFormFields function was because then the req.session.userCase.applicationType! couldn't be undefined, where as before it could be (for app1 answering the first couple of questions)

I'm happy to change it back if you'd prefer though

PR #1543, opened 2021-12-13
change the way fixture data is imported to the case

linusnorton:


src/test/steps/common.ts
```diff
@@ -268,42 +271,43 @@ export const iGetTheCaseApi = (testUser: UserDetails): CaseApi => {
...
+  const connections = addConnection(data);
 
-      await fetch(requestUrl, request);
-      await new Promise(resolve => setTimeout(resolve, 1500));
-      await fetch(redirectUrl, request);
-    },
-    [userCaseObj, requestPageLink, redirectPageLink]
-  );
+  // don't set as applicant 2 as they don't have permission
+  data.connections = connections.length > 0 ? connections : undefined;
```
velimd, line 295: would this still not set it for app2? i guess app2 will only use jointApp2 fixture and the data isnt enough?

velimd (approved):

src/test/steps/common.ts
linusnorton: yeah, it's not very robust but I couldn't think of an easy way to work out whether it's app two

PR #1617, opened 2022-01-07
Nfdiv 1555

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1555

### Change description ###

```
[ ] Yes
[X] No
```


src/main/routes.ts
```diff
@@ -60,12 +65,23 @@ export class Routes {
     app.post(DOCUMENT_MANAGER, handleUploads.array('files[]', 5), errorHandler(documentManagerController.post));
     app.get(`${DOCUMENT_MANAGER}/delete/:index`, errorHandler(documentManagerController.delete));
 
+    const isRouteForUser = (req: AppRequest, res: Response, next: NextFunction): void => {
+      const isApp2Route = [APPLICANT_2, RESPONDENT].some(prefixUrl => req.path.includes(prefixUrl));
+      if ((isApp2Route && !req.session.isApplicant2) || (!isApp2Route && req.session.isApplicant2)) {
+        return res.redirect('/');
```
linusnorton, line 71: infinite death loop?

src/main/routes.ts
```diff
@@ -81,6 +97,12 @@ export class Routes {
       errorHandler(new AccessCodePostController(applicant2AccessCodeContent.form.fields).post)
     );
 
+    app.get(SWITCH_TO_SOLE_APPLICATION, errorHandler(new SwitchToSoleApplicationGetController().get));
+    app.post(
+      SWITCH_TO_SOLE_APPLICATION,
+      errorHandler(new SwitchToSoleApplicationPostController(switchToSoleAppContent.form.fields).post)
+    );
```
linusnorton, line 104: Why were these two needed?

linusnorton: I see

src/main/steps/switch-to-sole-application/get.ts
```diff
@@ -0,0 +1,9 @@
+import { GetController } from '../../app/controller/GetController';
+
+import { generateContent } from './content';
+
+export class SwitchToSoleApplicationGetController extends GetController {
```
linusnorton, line 5: why is this needed?

src/test/functional/features/applicant-middleware.feature
```diff
@@ -0,0 +1,21 @@
+Feature: Homepage
+
+  Background:
```
linusnorton, line 3: feels like this should be a nightly test

src/main/routes.ts
velimd: dont know what that is, sounds like an anime?

src/main/steps/switch-to-sole-application/get.ts
velimd: i had to remove this from app1 sequence as app2 also uses the page, i guess i could have just used a normal get controller but i noticed every other page outside of the applicant routes were uses custom controllers that did the same thing.

src/test/functional/features/applicant-middleware.feature
velimd: that can be arranged

src/main/routes.ts
linusnorton: what if you get redirected to / and that in turn redirects you to /applicant2/ something, which triggers this and redirects you back to / - it might be best to redirect to the error page.

linusnorton (approved):

lara-mpt (approved):

PR #1618, opened 2022-01-07
Nfdiv 1207

velimd:
### JIRA link ###

https://tools.hmcts.net/jira/browse/NFDIV-1207

### Change description ###

Note: was waiting on the url for the new link but for now itll use money and property until content can find it.

```
[ ] Yes
[X] No
```


src/main/steps/applicant1/hub-page/template.njk
```diff
@@ -32,6 +32,9 @@
 
         <h3 class="govuk-heading-s">{{ subHeading2 }}</h3>
         <p class="govuk-body">{{ line1 | safe }}</p>
+        {% if progressionIndex == 11 %}
+          <p class="govuk-body">{{ line2 | safe }}</p>
+        {% endif %}
```
lara-mpt, line 37: I asked Sanchita about these links and she said for now we can leave those out

src/main/steps/home/get.ts
```diff
@@ -86,16 +86,11 @@ const applicant1RedirectPageSwitch = (caseState: State, userCase: Partial<Case>,
...
-    case State.AosOverdue:
-    case State.Holding:
-    case State.AwaitingGeneralConsideration: {
-      return HUB_PAGE;
+    case State.Draft: {
+      return isFirstQuestionComplete ? CHECK_ANSWERS_URL : YOUR_DETAILS_URL;
     }
     default: {
-      return isFirstQuestionComplete ? CHECK_ANSWERS_URL : YOUR_DETAILS_URL;
+      return HUB_PAGE;
```
lara-mpt, line 93: does this mean we can get rid of some of the get.tests, or is it worth still keeping them?

lara-mpt (approved):

src/main/steps/applicant1/hub-page/template.njk
velimd: yh, i added the info to the 'Change description'

src/main/steps/home/get.ts
velimd: we can keep it,
just noticed how big the other condition was getting so i felt like default should be hub page as really there are only a few states for this which is really draft as the others are convered.

linusnorton (approved):

PR #1619, opened 2022-01-07
Nfdiv-1319 Sole hub page CO pronounced

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1319

### Change description ###
Nfdiv-1319 Sole hub page CO pronounced

```
[ ] Yes
[x] No
```


src/main/steps/applicant1/hub-page/right-column/content.ts
```diff
@@ -4,29 +4,42 @@ import { CommonContent } from '../../../common/common.content';
...
-  } service (PDF)`,
-  deemedOrDispensedDownloadReference: `/downloads/${
-    userCase.alternativeServiceOutcomes?.[0].value.alternativeServiceType === AlternativeServiceType.DISPENSED
-      ? 'certificate-of-dispense-with-service'
-      : 'certificate-of-deemed-as-service'
-  }`,
+  applicationDownload: {
+    name: 'Divorce-Application',
+    reference: `/downloads/${isDivorce ? 'divorce-application' : 'application-to-end-civil-partnership'}`,
+    link: `View the ${isDivorce ? 'divorce application' : 'application to end your civil partnership'} (PDF)`,
```
velimd, line 10: maybe this should be called text and reference should be called link? im happy for reference to stay but link is more like text.

velimd: kinda funny how name is more like reference (of the download file) and reference is more like link and link is more like text. if this makes sense.

src/main/steps/applicant1/hub-page/right-column/content.ts
```diff
@@ -4,29 +4,42 @@ import { CommonContent } from '../../../common/common.content';
...
+      userCase.alternativeServiceOutcomes?.[0].value.alternativeServiceType === AlternativeServiceType.DISPENSED
+        ? 'dispensed'
+        : 'deemed'
+    } service (PDF)`,
+  },
+  certificateOfEntitlementDownload: {
+    reference: '/downloads/certificate-of-service',
+    name: 'Certificate-of-Entitlement',
+    link: 'View the certificate of entitlement (PDF)',
+  },
```
velimd, line 42: was there a specific reason for this? i feel like it became unnecessarily long

src/main/app/case/from-api-format.ts
```diff
@@ -97,6 +97,12 @@ const fields: FromApiConverters = {
   dueDate: data => ({
     dueDate: dayjs(data.dueDate).format('D MMMM YYYY'),
   }),
+  coDateAndTimeOfHearing: data => ({
+    coDateAndTimeOfHearing: dayjs(data.coDateAndTimeOfHearing).format('D MMMM YYYY'),
```
velimd, line 101: i split this into two variables as this value has both date and time. [pr](https://github.com/hmcts/nfdiv-frontend/pull/1631/files#diff-f11587349bf8795d977c6abbc6f3877355a7f79117b38a04085342e17aaeac35R114).

i guess its fine for you as you dont need it and i can just update it in my pr.

src/main/steps/applicant1/hub-page/sole/hub.njk
```diff
@@ -12,7 +12,17 @@
...
+    <p class="govuk-body">{{ conditionalOrderPronounced.line1 }}</p>
+    <p class="govuk-inset-text">{{ conditionalOrderPronounced.line2 }}</p>
+    <p class="govuk-body">{{ conditionalOrderPronounced.line3 }}</p>
+    <p class="govuk-body">{{ conditionalOrderPronounced.line4 }}
+      <a class="govuk-link" href={{conditionalOrderPronounced.downloadReference}}
+      download="Certificate-of-Entitlement">
+      {{ conditionalOrderPronounced.line5 }}</a>
+      {{ conditionalOrderPronounced.line6 }}
+    </p>
+{% elif progressionIndex == 11 %}
```
velimd, line 25: you added two new states but wouldnt this effect the current index order to display?

src/main/app/case/from-api-format.ts
lara-mpt: it worked just formatting it like this, is the time of the hearing something you need to retain?

lara-mpt: ah just saw it is on your pr, probably easier if you just update it with your pr after

src/main/steps/applicant1/hub-page/right-column/content.ts
lara-mpt: yeah you're right, I'll update those

src/main/steps/applicant1/hub-page/right-column/content.ts
lara-mpt: I remember an older pr where Taleb commented it would be better to move the html into the template, so I was updating them all to follow suit. It's longer but cleaner imo

velimd: i think he was talking about when we had massive html in one line but anchor tags were fine. Tbh i thought the other way was cleaner/easy to read but i understand your opinion as well that i dont mind. might be worth getting another eye on it.

lara-mpt: likewise, I'm happy either way but better to be consistent with it

src/main/steps/respondent/hub-page/template.njk
```diff
@@ -3,7 +3,16 @@
...
+{% if progressionIndex == 15 %}
+    <p class="govuk-body">{{ conditionalOrderPronounced.line1 }}</p>
+    <p class="govuk-inset-text">{{ conditionalOrderPronounced.line2 }}</p>
+    <p class="govuk-body">{{ conditionalOrderPronounced.line3 }}</p>
+    <p class="govuk-body">{{ conditionalOrderPronounced.line4 }}
+      <a class="govuk-link" href={{conditionalOrderPronounced.downloadReference}}
+      download="Certificate-of-Entitlement">
+      {{conditionalOrderPronounced.line5}}</a>
+    </p>
+  {% elif progressionIndex == 11 %}
```
velimd, line 15: i missed this as well but this page will also need to be considered due to the added states.

src/main/steps/applicant1/hub-page/right-column/content.ts
benouaer: I think it's much cleaner this way. now it's all data/variables here, which then gets used in the template to create the markup.

src/main/steps/applicant1/hub-page/sole/content.ts
```diff
@@ -146,6 +146,27 @@ const en = ({ isDivorce, partner, userCase }: CommonContent) => ({
   moneyAndProperty: `You can use the time to decide how your money and property will be divided. This is dealt with separately to the ${
     isDivorce ? 'divorce application' : 'application to end your civil partnership'
   }. <a class="govuk-link" href="https://www.gov.uk/money-property-when-relationship-ends" target="_blank">Find out about dividing money and property</a>`,
```
benouaer, line 148: is this an exception?

src/main/steps/applicant1/hub-page/progress-bar.njk
```diff
@@ -55,20 +55,20 @@
     label: {
       text: conditionalOrderGranted
     },
-    complete: progressionIndex > 11,
+    complete: progressionIndex > 12,
```
benouaer, line 58: these progression indices look a bit like magic numbers. I haven't looked at the code so not sure how they're used. but it seems dangerous to rely on an index in this way. is it not possible to use an ordered list

benouaer (approved):

src/main/steps/applicant1/hub-page/sole/content.ts
lara-mpt: @benouaer I can make a separate pr after to clean up existing links, as there are quite a few scattered about

src/main/steps/applicant1/hub-page/progress-bar.njk
lara-mpt: https://github.com/hmcts/nfdiv-frontend/blob/master/src/main/steps/applicant1/hub-page/sole/content.ts#L160-#L175

In terms of how the page content shows up it could probably be changed, but I can't think of a better way of implementing it for the progress bar.

 @velimd any opinions?

velimd: We can just use the state names. the idea was to use the index of states as the states model changed a few times at the time so we didnt have to rely on it.

Not sure what you mean by using a ordered list?

src/main/steps/applicant1/hub-page/sole/content.ts
velimd: theres going to be a few where the link is mid sentence, this is why i thought anchor tags were fine to use in the content file.

lara-mpt: a mid sentence link could become quite messy to separate out

velimd: exactly, though if both are you think this is cleaner then we can change it and i guess we can have an exception for links in sentences.

src/main/steps/applicant1/hub-page/progress-bar.njk
benouaer: what I meant by ordered list was just a sequence, like we have the sequence of the pages. so you just pick the next item in the sequence, if you don't want to mention the name of the state explicitly

src/main/steps/applicant1/hub-page/sole/content.ts
benouaer: we can make an exception for mid-sentence links
the thing with these anchor tags, is that they're not like a simple <br> or <p> tag, where you just have a start and end tag. you've got the class, href and target attributes, so it feels like it belongs to a template file

benouaer: it doesn't have to be done in this PR, we can do it gradually as needed

velimd (approved):

PR #1624, opened 2022-01-10
NFDIV-1204 & NFDIV-1176

tomxelliott:
Resolve issues related to AC3 in NFDIV-1204 and AC5 in NFDIV-1176

### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1204
https://tools.hmcts.net/jira/browse/NFDIV-1176

Screenshot 2022-01-11 at 10 04 26

Screenshot 2022-01-11 at 10 04 14


src/main/steps/index.ts
```diff
@@ -68,6 +68,15 @@ export const getNextIncompleteStepUrl = (req: AppRequest): string => {
   return `${url}${queryString}`;
 };
 
+export const getNextIncompleteConditionalOrderUrl = (req: AppRequest): string => {
```
linusnorton, line 71: maybe this should be somewhere else, or more generic:

getNextIncompletePageAfter(req, page)

src/main/steps/index.ts
```diff
@@ -63,7 +63,13 @@ const getNextIncompleteStep = (
 export const getNextIncompleteStepUrl = (req: AppRequest): string => {
   const { queryString } = getPathAndQueryString(req);
   const sequence = getUserSequence(req);
-  const url = getNextIncompleteStep(req.session.userCase, sequence[0], sequence, true);
+  const sequenceIndex =
+    !req.session.isApplicant2 &&
+    (req.session.userCase.state === State.ConditionalOrderDrafted ||
```
velimd, line 68: might be cleaner to do [State.ConditionalOrderDrafted, State.ConditionalOrderPending].includes(req.session.userCase.state)

src/main/steps/index.ts
```diff
@@ -63,7 +63,13 @@ const getNextIncompleteStep = (
...
+      req.session.userCase.state === State.ConditionalOrderPending)
+      ? sequence.findIndex(s => s.url.includes(READ_THE_RESPONSE))
+      : 0;
```
velimd, line 71: is this because theres a break in the sequence?

i think if thats the case you can move the urls closer to match the actual sequence as i know sometimes we just add it to the end which might break this and this will be unnecessary.

velimd (approved):

PR #1662, opened 2022-01-19
Nfdiv-1157 CYA page for joint CO

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1157

### Change description ###
 CYA page for joint CO

```
[ ] Yes
[x] No
```


src/main/steps/index.ts
```diff
@@ -52,6 +52,8 @@ const getNextIncompleteStep = (
 
       return nextStep
         ? getNextIncompleteStep(data, nextStep, sequence, removeExcluded, checkedSteps.concat(step))
+        : [State.ConditionalOrderDrafted, State.ConditionalOrderPending].includes(data.state)
```
velimd, line 55: i think this needs to be its own if/case statement.

src/main/steps/applicant1/check-your-conditional-order-answers/template.njk
```diff
@@ -6,6 +6,7 @@
 {% set nextIncompleteStep = getNextIncompleteStepUrl() %}
 {% set isConditionalOrderAnswersComplete =
         '/check-your-conditional-order-answers' in nextIncompleteStep or
+        nextIncompleteStep === '/applicant2/' or
```
velimd, line 9: is there a page which is just '/applicant2/'?

src/main/steps/index.ts
lara-mpt: as in split that return statement into multiple if/else or that the check for that should be separate from the nextStep check? 

I thought it was worth keeping it in the nextStep check to make sure the continue-with-your-application question had been answered

src/main/steps/applicant1/check-your-conditional-order-answers/template.njk
lara-mpt: There isn't but I double checked and this is needed for app2 as the getNextIncompleteStepUrl returns '/applicant2' for them when they finish the sequence

src/main/steps/index.ts
velimd: just make it 1 if/else statement as i feel like the inline condition should really be used for if or else but when theres multiple conditions it should just be its own if (nextstep) else if (this state) else ....

src/main/steps/applicant1/check-your-conditional-order-answers/content.ts
```diff
@@ -16,12 +16,20 @@ const en = ({ isJointApplication, isDivorce, userCase }: CommonContent) => ({
     isInformationCorrect: 'Is the information in this application still correct?',
   },
   stepAnswers: {
-    continueApplication: userCase.applicant1ApplyForConditionalOrder,
-    isInformationCorrect: userCase.applicant1ConfirmInformationStillCorrect,
+    continueApplication: `${
```
velimd, line 19: how comes here its continueApplication and actually property names and in check your answers its line1, line2...?

velimd (approved):

src/main/steps/applicant1/check-your-conditional-order-answers/content.ts
lara-mpt: @velimd it was previously called continueApplication and isInformationCorrect, I can update it to line1 etc to be more consistent if you'd prefer?

velimd: na its kl, was just curious hence i still approved.

PR #1691, opened 2022-01-24
Nfdiv 1546

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1546

### Change description ###
image

```
[ ] Yes
[X] No
```


src/main/app/case/case.ts
```diff
@@ -73,7 +74,9 @@ export const formFieldsToCaseMapping: Partial<Record<keyof Case, keyof CaseData>
   applicant2LegalProceedings: 'applicant2LegalProceedings',
   applicant2LegalProceedingsDetails: 'applicant2LegalProceedingsDetails',
   applyForFinancialOrder: 'applicant1FinancialOrder',
+  whoIsFinancialOrderFor: 'applicant1FinancialOrderFor',
```
lara-mpt, line 77: can't we change these to _applicant1WhoIsFinancialOrderFor_ and _applicant1WhoIsFinancialOrderFor_ for consistency now?

src/main/app/case/to-api-format.ts
```diff
@@ -218,6 +218,13 @@ const fields: ToApiConverters = {
   coApplicant2StatementOfTruth: data => ({
     coApplicant2StatementOfTruth: checkboxConverter(data.coApplicant2StatementOfTruth),
   }),
+  whoIsFinancialOrderFor: data => ({
+    applicant1FinancialOrderFor: data.applyForFinancialOrder !== YesOrNo.NO ? data.whoIsFinancialOrderFor : [],
```
lara-mpt, line 222: imo it's better to use === YesOrNo.Yes since if applyForFinancialOrder is undefined we wouldn't want whoIsFinancialOrderFor to be set anyway

src/main/steps/applicant1/check-your-answers/content.ts
```diff
@@ -547,6 +557,7 @@ const cy: typeof en = ({
     },
     dividingAssets: {
       line1: 'Do you want to apply for a financial order?',
+      line2: 'Who is the financial for?',
```
lara-mpt, line 560: ignore this if it's right, but looks like it's missing an 'order' and should be 'Who is the financial order for?'

src/main/steps/applicant1/confirm-your-joint-application/content.ts
```diff
@@ -91,8 +92,25 @@ const en = ({ isDivorce, partner, userCase, userEmail, isApplicant2 }: CommonCon
...
+    isEqual(userCase.whoIsFinancialOrderFor, [FinancialOrderFor.APPLICANT, FinancialOrderFor.CHILDREN]) ? ' and ' : ''
+  }${
+    userCase.whoIsFinancialOrderFor?.includes(FinancialOrderFor.CHILDREN)
+      ? 'for the children of the applicant and the respondent.'
+      : '.'
+  }`,
+  applicant2FinancialOrderYes: `Applicant 2 is applying to the court for financial orders ${
+    userCase.applicant2WhoIsFinancialOrderFor?.includes(FinancialOrderFor.APPLICANT) ? 'for the applicant' : ''
+  }${
+    isEqual(userCase.applicant2WhoIsFinancialOrderFor, [FinancialOrderFor.APPLICANT, FinancialOrderFor.CHILDREN])
```
danielbe11, line 107: Can we guarantee that [FinancialOrderFor.APPLICANT, FinancialOrderFor.CHILDREN] will always be in the same order? cos isEqual will return false for [FinancialOrderFor.CHILDREN, FinancialOrderFor.APPLICANT]

velimd: for some reason using sort broke the comparison before. tho now im using lodash so it might work and ill add it in.

src/main/steps/applicant1/confirm-your-joint-application/content.ts
```diff
@@ -91,8 +92,25 @@ const en = ({ isDivorce, partner, userCase, userEmail, isApplicant2 }: CommonCon
...
-  applicant2FinancialOrderYes: 'Applicant 2 is applying to the court for financial orders',
+  applicant1FinancialOrderYes: `Applicant 1 is intending to apply to the court for financial orders ${
+    userCase.whoIsFinancialOrderFor?.includes(FinancialOrderFor.APPLICANT) ? 'for the applicant' : ''
+  }${
+    isEqual(userCase.whoIsFinancialOrderFor, [FinancialOrderFor.APPLICANT, FinancialOrderFor.CHILDREN]) ? ' and ' : ''
+  }${
+    userCase.whoIsFinancialOrderFor?.includes(FinancialOrderFor.CHILDREN)
+      ? 'for the children of the applicant and the respondent.'
+      : '.'
+  }`,
```
lara-mpt, line 103: haven't tried so not 100% sure but we could do 

`userCase.whoIsFinancialOrderFor?.join(' and ').replace(FinancialOrderFor.APPLICANT, 'for the applicant').replace(FinancialOrderFor.CHILDREN, 'for the children of the applicant and the respondent')`

lara-mpt: also this is the confirm-your-joint-application page so I assume 'respondent' should be 'applicant 2' and 'applicant' should be 'applicant 1'

src/main/app/case/to-api-format.ts
velimd: it wouldnt get called unless the field whoIsFinancialOrderFor is being passed as a property to send to the backend which is always tagged with applyForFinancialOrder as its on the same page. im happy to change it tho.

src/main/steps/applicant1/check-your-answers/content.ts
velimd: yep my bad, i wrote the question wrong and copied it from there.

src/main/app/case/to-api-format.ts
lara-mpt: yeah fair, I guess it doesn't matter either way then. I think === looks nicer, but I don't mind since it's just preference

src/main/steps/applicant1/confirm-your-joint-application/content.ts
velimd: that method is sooo much better, nice one.

src/main/app/case/to-api-format.test.ts
```diff
@@ -148,6 +153,10 @@ describe('to-api-format', () => {
       applicant2NameChangedHow: [],
       applicant1CannotUploadSupportingDocument: [DocumentType.NAME_CHANGE_EVIDENCE],
       applicant2CannotUploadSupportingDocument: [DocumentType.NAME_CHANGE_EVIDENCE],
+      applicant1FinancialOrder: YesOrNo.YES,
+      applicant1FinancialOrderFor: [FinancialOrderFor.APPLICANT, FinancialOrderFor.CHILDREN],
+      applicant2FinancialOrder: YesOrNo.NO,
+      applicant2FinancialOrderFor: [],
```
lara-mpt, line 159: just for coverage sake it's worth having a test in 'results' with applicant1FinancialOrder: YesOrNo.NO, and applicant1FinancialOrderFor: [FinancialOrderFor.APPLICANT, FinancialOrderFor.CHILDREN] to test the unsetting of the applicant1FinancialOrderFor field

and same for app2 fields

lara-mpt (approved):

src/main/app/case/to-api-format.test.ts
velimd: done

PR #1716, opened 2022-02-01
Nfdiv 1546

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1546

### Change description ###
NFDIV-1546

```
[ ] Yes
[X] No
```


velimd:
had to re open due to failing idam test but other prs were fine.

src/main/app/case/to-api-format.test.ts
```diff
@@ -333,4 +342,57 @@ describe('to-api-format', () => {
...
+      applicant1WhoIsFinancialOrderFor: [FinancialOrderFor.APPLICANT, FinancialOrderFor.CHILDREN],
+      applicant2ApplyForFinancialOrder: YesOrNo.NO,
+      applicant2WhoIsFinancialOrderFor: [FinancialOrderFor.APPLICANT, FinancialOrderFor.CHILDREN],
+      expected: {
+        applicant1FinancialOrder: YesOrNo.YES,
+        applicant1FinancialOrdersFor: [FinancialOrderFor.APPLICANT, FinancialOrderFor.CHILDREN],
+        applicant2FinancialOrder: YesOrNo.NO,
+        applicant2FinancialOrdersFor: [],
+      },
+    },
```
lara-mpt, line 394: applicant1ApplyForFinancialOrder and applicant2ApplyForFinancialOrder are unrelated so how comes your testing Yes/No and No/Yes situations? I would have thought Yes Yes and No No would be enough

src/test/functional/features/form-failures.feature
```diff
@@ -223,6 +223,10 @@ Feature: Form failures
     When I click "Continue"
     Then the page should include "You have not answered the question. You need to select an answer before continuing."
 
+    Given I select "Yes, I want to apply for a financial order"
+    When I click "Continue"
+    Then the page should include "You have not answered the question. You need to select an answer before continuing."
```
lara-mpt, line 228: could include this on the block above since it's the same page

lara-mpt (approved):

src/main/app/case/to-api-format.test.ts
velimd: in the to-api-format, if they select no, it sets the array to empty, hence im doing different situations.

lara-mpt: but you're already testing that no situation in the test above?

velimd: just thought ill add extra test to indicate one doesnt effect the other.
so saying no to one doesnt set the other applicants array.
i can remove it if you prefer, i was just covering different situations.

lara-mpt: yeah seems a bit unnecessary to me

PR #1765, opened 2022-02-15
stop sending data when triggering switch-to-sole event

benouaer:
### JIRA link ###
- https://tools.hmcts.net/jira/browse/NFDIV-1707

### Change description ###
- stop sending data when triggering switch-to-sole event

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/steps/applicant1/application-ended/get.ts
+++ b/src/main/steps/applicant1/application-ended/get.ts
@@ -1,7 +1,6 @@
 import autobind from 'autobind-decorator';
 import { Response } from 'express';
 
-import { getSystemUser } from '../../../app/auth/user/oidc';
 import { getCaseApi } from '../../../app/case/CaseApi';
 import { ApplicationType, SWITCH_TO_SOLE } from '../../../app/case/definition';
 import { AppRequest } from '../../../app/controller/AppRequest';
@@ -16,8 +15,7 @@ export default class ApplicationEndedGetController extends GetController {
   }
 
   public async get(req: AppRequest, res: Response): Promise<void> {
-    const caseworkerUser = await getSystemUser();
-    req.locals.api = getCaseApi(caseworkerUser, req.locals.logger);
+    req.locals.api = getCaseApi(req.session.user, req.locals.logger);
 
     try {
       req.session.userCase.applicationType = ApplicationType.SOLE_APPLICATION;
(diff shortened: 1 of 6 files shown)


src/main/steps/applicant1/application-ended/get.ts
```diff
@@ -16,8 +15,7 @@ export default class ApplicationEndedGetController extends GetController {
   }
 
   public async get(req: AppRequest, res: Response): Promise<void> {
-    const caseworkerUser = await getSystemUser();
-    req.locals.api = getCaseApi(caseworkerUser, req.locals.logger);
+    req.locals.api = getCaseApi(req.session.user, req.locals.logger);
```
lara-mpt, line 18: I think this would be set already so you don't need to specify it here

src/test/functional/features/switch-to-sole-application.feature
```diff
@@ -7,7 +7,7 @@ Feature: Switch To Sole Application
     And I click "Send for review"
     Then the page URL should be "/application-sent-for-review"
 
-  Scenario: Switching to Sole application by Applicant 1
+  Scenario: Switching to Sole application by Applicant one
```
lara-mpt, line 10: we use applicant 1 and applicant 2 everywhere else, so do we need to change this?

benouaer: I just did that here because when you try to run one test with grep, it picks up all the other tests in the feature which have applicant 1. a better solution would be for me to number the scenarios, let me do that instead

src/test/functional/features/switch-to-sole-application.feature
```diff
@@ -40,15 +40,15 @@ Feature: Switch To Sole Application
...
+  Scenario: [2] Switching to Sole application by Applicant 1 before Applicant 2 links to the case
     Given I go to '/switch-to-sole-application'
     When I click "Create a new application"
     Then the page URL should be "/your-details"
     Given I go to "/"
     Then the page URL should be "/check-your-answers"
 
 
-  Scenario: Switching to Sole application by Applicant 2
+  Scenario: [3] Switching to Sole application by Applicant two
```
lara-mpt, line 51: missed a `two`

lara-mpt (approved):

PR #1788, opened 2022-02-22
no response page moved to parent dir due to common use.

velimd:
### Change description ###
image

```
[ ] Yes
[X] No
```


src/main/steps/no-response-yet/content.ts
```diff
@@ -0,0 +1,22 @@
+import { TranslationFn } from '../../app/controller/GetController';
+import { SWITCH_TO_SOLE_APPLICATION } from '../urls';
+
+const en = ({ partner }) => ({
+  title: `Your ${partner} has not yet responded`,
+  line1: `Your ${partner} has not yet responded. The quickest way to progress your application is to contact them and ask them to respond, if it’s safe to do so. They can still respond, even though the deadline has passed.`,
+  readMore: 'If you do not think they will respond',
+  line2: `If you think that your ${partner} will not respond then you can create a new ‘sole application’.`,
+  line3: `This means that you will submit the application on your own, and your ${partner} will respond to it after it has been submitted and checked by the court.`,
+  line4: `You can <a href="${SWITCH_TO_SOLE_APPLICATION}" class="govuk-link">create a new application here.</a>`,
```
benouaer, line 10: could the anchor tag be moved to the template?

src/main/steps/no-response-yet/get.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  test('Should render the no response page with civil content', async () => {
+    const req = mockRequest();
+    const res = mockResponse();
+    res.locals.serviceType = DivorceOrDissolution.DISSOLUTION;
+    await controller.get(req, res);
+    const isDivorce = false;
+
+    expect(res.render).toBeCalledWith(expect.anything(), {
```
benouaer, line 40: is there a way to check the actual text displayed on the page?

src/main/steps/no-response-yet/get.ts
```diff
@@ -0,0 +1,9 @@
+import { GetController } from '../../app/controller/GetController';
+
+import { generateContent } from './content';
+
+export class NoResponseYetApplicationGetController extends GetController {
+  constructor() {
+    super(__dirname + '/template.njk', generateContent);
```
benouaer, line 7: not sure what's happening here

benouaer (approved):

src/main/steps/no-response-yet/get.ts
velimd: its just setting the controller up with the template location and content function. its either set it here or set it in routes.ts but having to import every content function and set the path for every page that isnt in the sequence file.

src/main/steps/no-response-yet/get.test.ts
velimd: we can match some content

velimd: the expect anything part is just the path to the template file, do we need to test that?

PR #1800, opened 2022-02-28
NFDIV-1765 sanitize user input

linusnorton:
Remove tags from input fields when displaying the Check Your Answers page.

Ideally we would remove | safe from the nunjucks template but there is a lot of embedded HTML.


src/main/steps/applicant1/check-your-answers/content.ts
```diff
@@ -654,8 +658,8 @@ const cy: typeof en = ({
       line1: `${userCase.inTheUk.replace('Yes', 'Do').replace('No', 'Naddo')}`,
```
danielbe11, line 658: Can we use stripTags here too?

linusnorton: I didn't think it was necessary for YesOrNo fields

danielbe11: Ah yeah, I was just comparing to L225 but I agree it's not needed.

danielbe11 (approved):

PR #1809, opened 2022-03-01
replace progressIndex arrays with StateSequence class for better readability

benouaer:
### Change description ###
- replace progressIndex arrays with StateSequence class for better readability

```
[ ] Yes
[X] No
```


linusnorton (approved):
I hadn't seen StateSequence before. I'm not sure I like the idea as the states aren't really linear but the PR looks fine.

PR #1813, opened 2022-03-02
Nfdiv-1507 Update jurisdiction

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1507

### Change description ###
Update jurisdiction and clean up

```
[ ] Yes
[x] No
```


src/main/app/jurisdiction/connections.ts
```diff
@@ -92,7 +93,19 @@ const hasResidualJurisdiction = (data, connections) => {
   );
 };
 
-export const addConnection = (data: Partial<CaseWithId>): JurisdictionConnections[] => {
+export const addConnections = (reqUrl: string, data: Partial<CaseWithId>): JurisdictionConnections[] => {
```
linusnorton, line 96: I'm not sure I like this approach. It feels like we're cramming situation specific logic into the wrong place. My gut feel is that the calling method should know whether or not it should be behave this way. I don't have a full understanding but I would suggest the calling code that needs both variations of this method should be something like `addConnectionsBasedOnQuestions().concat(addConnections())`

linusnorton (approved):
I think the addConnections change could do with a re-think but if you disagree then it's fine

linusnorton (approved):

src/main/app/jurisdiction/moreDetailsContent.ts
```diff
@@ -0,0 +1,75 @@
...
+  const connectionTypes = {
+    'Habitual residence': [
+      JurisdictionConnections.APP_1_APP_2_LAST_RESIDENT,
+      JurisdictionConnections.APP_2_RESIDENT,
+      JurisdictionConnections.APP_1_RESIDENT_SIX_MONTHS,
+      JurisdictionConnections.APP_1_APP_2_RESIDENT,
+      JurisdictionConnections.APP_1_RESIDENT_JOINT,
+      JurisdictionConnections.APP_1_RESIDENT_TWELVE_MONTHS,
+    ],
+    Domicile: [
```
danielbe11, line 39: Why no '' unlike L31?

lara-mpt: because it's only 1 word it didn't need it. There's a reason why for its use in the connectionText but can't remember specifically

PR #1852, opened 2022-03-10
NFDIV 333 - Add Webchat

danielbe11:
### JIRA link ###

https://tools.hmcts.net/jira/browse/NFDIV-333

### Change description ###

- Adding Webchat.

```
[ ] Yes
[X] No
```


src/main/modules/nunjucks/index.ts
```diff
@@ -75,6 +77,16 @@ export class Nunjucks {
       }));
     });
 
+    const globals = {
+      webchat: {
+        avayaUrl: config.webchat.avayaUrl,
```
velimd, line 82: just for consistency, we should use the get method of config.

src/main/steps/common/common.content.ts
```diff
@@ -98,6 +98,28 @@ const en = {
   apmCookiesHeadings: 'Allow cookies that measure website application performance monitoring?',
   useApmCookies: 'Use cookies that measure website application performance monitoring',
   doNotUseApmCookies: 'Do not use cookies that measure website application performance monitoring',
+  helpTitle: 'Contact us for help',
+  helpHeading1: 'Telephone',
+  helpHeading2: 'Email',
+  helpHeading3: 'Webchat',
+  helpEmailLabel:
+    'Email us at <a href=\'mailto:{contactEmailAddress}\' class="govuk-link" tabindex="-1">{contactEmailAddress}</a>.',
```
velimd, line 106: Taleb will not like this.

src/main/steps/common/page.njk
```diff
@@ -112,4 +112,6 @@
     })(window,document,'script','dataLayer','GTM-WG59GG2');</script>
   <!-- End Google Tag Manager -->
 
+  <script src="https://{{ globals.webchat.avayaClientUrl }}/chat-client/1/app-es2015.js" defer type="module" crossorigin="anonymous"></script>
```
velimd, line 115: what does this script do as we have a webchat script already, curious what the difference is.

src/main/assets/locales/avaya-webchat/locale_en-GB.json
```diff
@@ -0,0 +1,64 @@
...
+      "no": "No"
+    },
+    "offerSurvey": "Send me a feedback survey"
+  },
+  "survey": {
+    "propose": "Would you like to take part in an email survey at the end of your Webchat?",
+    "yes": "Yes",
+    "no": "No"
+  }
+}
```
velimd, line 64: i dont see where this is used or how its used, i assumed the other script.

src/main/steps/common/page.njk
danielbe11: The avaya-webchat.ts code determines whether to show the link to the webchat or not e.g. if webchat is down, the link will not be shown and some info will be shown instead. app-es2015.js is the code for the actual webchat.

src/main/assets/locales/avaya-webchat/locale_en-GB.json
danielbe11: Yeah this is used in the other script. If you're testing it, it's likely that none of the webchat agents are available (cos its using pre prod URL) and the only content you'll see rendered is L4 "No Advisors are currently available"

lara-mpt (approved):

PR #1862, opened 2022-03-13
Refactor applicant1/you-can-use-english-welsh-courts/content.ts

danielbe11:
Incorporates #1859

### DIFF
--- a/src/main/steps/applicant1/you-can-use-english-welsh-courts/content.ts
+++ b/src/main/steps/applicant1/you-can-use-english-welsh-courts/content.ts
@@ -88,7 +88,7 @@ export const form: FormContent = {
     const removeConnectionJ = c => c !== JurisdictionConnections.APP_1_RESIDENT_JOINT;
     const removeConnectionK = c => c !== JurisdictionConnections.APP_2_RESIDENT_TWELVE_MONTHS;
     const removeConnectionL = c => c !== JurisdictionConnections.APP_2_RESIDENT_SIX_MONTHS;
-    const removeConnectionIIfOppositeSexDivorce = c => c !== JurisdictionConnections.RESIDUAL_JURISDICTION;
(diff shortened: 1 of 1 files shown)


src/main/steps/applicant1/you-can-use-english-welsh-courts/content.ts
```diff
@@ -75,29 +76,41 @@ const cy = en;
 
 export const form: FormContent = {
   fields: userCase => {
+    // Create map of connections to labels:
+    const connectionsLabelMap = new Map<JurisdictionConnections, Label>();
+    jurisdictionConnectionList.forEach((key: JurisdictionConnections, i: number) =>
+      connectionsLabelMap.set(key, l => l.connectionCheckboxes[i])
```
lara-mpt, line 82: probably better to use 'index' or another name for I so that it doesn't get confused with connection I

src/main/steps/applicant1/you-can-use-english-welsh-courts/content.ts
```diff
@@ -75,29 +76,41 @@ const cy = en;
...
+    const removeConnectionK = c => c !== JurisdictionConnections.APP_2_RESIDENT_TWELVE_MONTHS;
+    const removeConnectionL = c => c !== JurisdictionConnections.APP_2_RESIDENT_SIX_MONTHS;
+    const removeConnectionI = c => c !== JurisdictionConnections.RESIDUAL_JURISDICTION;
+
+    const filters = [removePreMadeConditions];
+
+    if (userCase.applicationType === ApplicationType.SOLE_APPLICATION) {
+      filters.push(removeConnectionJ);
+      filters.push(removeConnectionK);
+      filters.push(removeConnectionL);
```
lara-mpt: would be cleaner if we can combine these removeConnections

lara-mpt (approved):

PR #1874, opened 2022-03-15
NFDIV-1879 Prevent users from updating case after submission

lara-mpt:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1879

### Change description ###
Prevent users from updating case after submission

```
[ ] Yes
[x] No
```


src/main/steps/index.ts
```diff
@@ -73,21 +91,45 @@ export const getNextIncompleteStepUrl = (req: AppRequest): string => {
...
+  const finalUrls = [HOME_URL, CHECK_ANSWERS_URL, `${APPLICANT_2 + CONFIRM_JOINT_APPLICATION}`];
+  const startsWithUrls = ['/pay', APPLICATION_SUBMITTED];
+
+  return (
+    finalUrls.some(url => url === nextStepUrl.split('?')[0]) || startsWithUrls.some(url => nextStepUrl.startsWith(url))
+  );
+};
+
+export const isConditionalOrderReadyToSubmit = (nextStepUrl: string): boolean => {
+  const finalUrls = [HOME_URL, CHECK_ANSWERS_URL, `${APPLICANT_2 + HOME_URL}`];
```
danielbe11, line 104:
```suggestion
  const finalUrls = [HOME_URL, `${APPLICANT_2 + HOME_URL}`];
```

danielbe11 (approved):

src/main/routes.ts
```diff
@@ -72,11 +72,16 @@ export class Routes {
 
     const isRouteForUser = (req: AppRequest, res: Response, next: NextFunction): void => {
       const isApp2Route = [APPLICANT_2, RESPONDENT].some(prefixUrl => req.path.includes(prefixUrl));
-      if ((isApp2Route && !req.session.isApplicant2) || (!isApp2Route && req.session.isApplicant2)) {
+      if (
+        (isApp2Route && !req.session.isApplicant2) ||
+        (!isApp2Route && req.session.isApplicant2) ||
```
linusnorton, line 77: I think these two lines are basically an XOR which would be logically equivalent to:

```
if (isApp2Route != req.session.isApplicant2 || !getUserSequence(req).find(r => req.path.includes(r.url))
```

I would also probably use `.some` rather than `.find` as you want a boolean back rather than the element itself but it works either way

src/main/steps/applicant1Sequence.test.ts
```diff
@@ -1,6 +1,6 @@
 import { ApplicationType } from '../app/case/definition';
 
-import { Step, applicant1Sequence, isCountryUk } from './applicant1Sequence';
+import { Step, applicant1PreSubmissionSequence, isCountryUk } from './applicant1Sequence';
```
linusnorton, line 3: if it's not too much trouble I would use the names:

```
applicant1ApplicationSequence
applicant2....
applicant1ConditionalOrderSequence
applicant1FinalOrderSequence
```

I'm not sure if we have a sequence for CO and FO but it would logically line up with the application phases if we did.

linusnorton: looking at the file below it seems like we just have pre and post submission so this comment might not be so relevant but I'll leave it up to you

linusnorton (approved):

PR #1899, opened 2022-03-21
Test Elastic Search 

linusnorton:

### Change description ###

```
[ ] Yes
[ ] No
```


src/test/steps/common.ts
```diff
@@ -322,6 +322,9 @@ const executeUserCaseScript = async (data, redirectPageLink: string) => {
   const user = testConfig.GetCurrentUser();
   const testUser = await iGetTheTestUser(user);
   const api = iGetTheCaseApi(testUser);
+
+  // add a delay after logging a user in because it creates and extra case that needs to be added to the ES index
```
benouaer, line 326: should we check for that case?

benouaer (approved):

src/test/steps/common.ts
```diff
@@ -322,6 +322,9 @@ const executeUserCaseScript = async (data, redirectPageLink: string) => {
...
+  await new Promise(resolve => setTimeout(resolve, 5000));
```
velimd, line 327: at what stage is this? wouldnt it be better ot make the above function return a promise and wait for it then do the next step?

linusnorton: replied on slack but that's because the log in step creates a case and we need to wait for it to be indexed

PR #1904, opened 2022-03-21
clean up cannot upload checkbox and cya page

velimd:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1753

### Change description ###
image
image

```
[ ] Yes
[X] No
```


src/main/steps/applicant1/check-your-answers/content.ts
```diff
@@ -57,9 +58,35 @@ const getOtherCourtCasesMoreDetailsContent = () => {
   return moreDetailsComponent({ text, title });
 };
 
+const cannotUploadDocumentList = (
```
benouaer, line 61: can we move this method to content.utils

velimd: its really only used here, do you think its worth moving? i think in the future these functions might just be incorporated with en/cy methods.

benouaer (approved):

lara-mpt (approved):

PR #1944, opened 2022-03-27
update prayer

linusnorton:
https://tools.hmcts.net/jira/browse/NFDIV-1931


velimd (approved):
LGTM, we already discuss the different outcomes, one thing that also popped in my head was maybe we could have done this in the events, if both citizen and xui do it, then it could have been a function in the event.

PR #2018, opened 2022-04-13
fix incorrect gender on aos page

aramchurn21:
### Change description ###

Enter a description.
Fixing incorrect gender on aos page

[jira link nfdiv-2136](https://tools.hmcts.net/jira/browse/NFDIV-2136)


src/main/steps/applicant1/review-the-application/content.ts
```diff
@@ -164,13 +164,22 @@ export const form: FormContent = {
...
 };
 
+const getApplicant1PartnerContent = (content: CommonContent): string => {
+  if (content.userCase?.sameSex !== Checkbox.Checked && content.partner !== content.civilPartner) {
+    return content.partner === content.husband ? content.wife : content.husband;
+  } else {
+    return content.partner;
+  }
+};
+
```
lara-mpt, line 174: since this is also used on the check-your-joint-application page it's probably worth moving the function into content.utils.ts so that we can reuse it

aramchurn21: Also updated check-your-joint-application page to include getApplicant1PartnerContent function from common content.utils.

src/main/steps/applicant1/review-the-application/content.ts
```diff
@@ -171,6 +172,7 @@ const languages = {
 
 export const generateContent: TranslationFn = (content: CommonContent) => {
   const { language, userCase } = content;
+  content.partner = getApplicant1PartnerContent(content);
```
lara-mpt, line 175: if you changed the partner here then all calls to ${partner} on the page will be referencing the Applicant1Partner which I don't think is correct. It's probably better to call this 

  `content.applicant1Partner = getApplicant1PartnerContent(content);`

and then update that specific partner mentioned in the ticket to be ${applicant1Partner}

aramchurn21: Changed the value of partner to applicant1Partner.

aramchurn21: All done.

lara-mpt (approved):

PR #2027, opened 2022-04-19
fix to use data to create form on next step sequence function

velimd:
### Change description ###

Enter a description.
Rework around getnextsteps functionalility due to upload your documents page field requires user data as its field is dynamic.

https://tools.hmcts.net/jira/browse/NFDIV-2129

- [X] tests have been updated / new tests has been added (if needed)


src/main/steps/index.ts
```diff
@@ -60,10 +60,12 @@ const getNextIncompleteStep = (
   removeExcluded = false,
   checkedSteps: Step[] = []
 ): string => {
-  const stepForm = stepForms[step.url];
+  const stepField = stepFields[step.url];
   // if this step has a form
-  if (stepForm !== undefined) {
+  if (stepField !== undefined) {
```
benouaer, line 65: does this have to be explicit, can it be: if (stepField)

benouaer (approved):

src/main/steps/index.ts
velimd: hmmm, im not sure, it could have been added because empty objects passed but i doubt it. ill have a look.

PR #2035, opened 2022-04-22
Improve and clean up functional tests

lara-mpt:
### Change description ###

Improve and clean up functional tests


src/test/steps/happy-path.ts
```diff
@@ -1,63 +1,44 @@
...
   const fixtureJson = require(`../functional/fixtures/${fixture}`)[fixture];
 
   await iSetTheUsersCaseTo(fixtureJson);
 
-  const url = await I.grabCurrentUrl();
-  I.amOnPage(APPLY_FINANCIAL_ORDER);
-  iClick('Continue');
-  I.amOnPage(url);
+  I.amOnPage(ADDRESS_PRIVATE);
+  I.click('I do not need my contact details kept private');
```
velimd, line 17: is there a reason for this?

i feel like the data should already exist and clicking continue should be enough.

velimd: same for below

velimd (approved):
LGTM just one comment

src/test/steps/happy-path.ts
lara-mpt: @velimd yes it's because when we update the case data using the iSetTheUsersCaseTo we need to then go to a page and click continue on it so that we fetch the data from ccd on the 'Continue'. Previously we were logging the user out and then back in but I thought it would save time this way. 

The reason I chose the ADDRESS_PRIVATE page to do it was because it's one of the few pages accessible for app1, app2 and the respondent

PR #2085, opened 2022-05-06
Upgrade redis to v4.1.0

lara-mpt:
### Change description ###

Upgrade redis to v4.1.0


src/main/modules/session/index.ts
```diff
@@ -37,16 +38,20 @@ export class SessionStorage {
   private getStore(app: Application) {
     const redisHost = config.get('session.redis.host');
     if (redisHost) {
-      const client = redis.createClient({
-        host: redisHost as string,
+      const redisClient = createClient({
+        socket: {
+          host: redisHost as string,
+          port: 6380,
```
benouaer, line 44: should we add this to config

src/main/modules/session/index.ts
```diff
@@ -37,16 +38,20 @@ export class SessionStorage {
...
+          tls: true,
+          connectTimeout: 15000,
```
benouaer, line 46: maybe this as well could be made a config variable

src/main/modules/session/index.ts
```diff
@@ -37,16 +38,20 @@ export class SessionStorage {
...
+        },
         password: config.get('session.redis.key') as string,
-        port: 6380,
-        tls: true,
-        connect_timeout: 15000,
       });
 
-      app.locals.redisClient = client;
-      return new RedisStore({ client });
+      redisClient.connect().then(() => {
```
benouaer, line 51: if possible, can we do this with async await

lara-mpt: I tried to do that yesterday but I don't think it's possible as it throws an error
`Top-level 'await' expressions are only allowed when the 'module' option is set to 'esnext'`

lara-mpt: or a similar error to that I can't remember exactly

benouaer (approved):

PR #2106, opened 2022-05-12
NFDIV-2302 - AOS Statement of truth set to YES when respondent submits AOS

danielbe11:
### Change description ###

- Currently the AcknowledgementOfService::statementOfTruth field isn't being set to Yes when citizen respondent submits their AOS response. This change does that.
- No validation required because of the existing validation for the applicant2IBelieveApplicationIsTrue field.

https://tools.hmcts.net/jira/browse/NFDIV-2216


velimd (approved):

src/main/steps/applicant2Sequence.ts
```diff
@@ -211,7 +211,7 @@ const postSubmissionSequence: Step[] = [
 const hasApp2Confirmed = (data: Partial<CaseWithId>): boolean =>
   ![State.AwaitingApplicant1Response, State.AwaitingApplicant2Response, State.Draft].includes(data.state as State) &&
   data.applicant2IConfirmPrayer === Checkbox.Checked &&
-  data.applicant2StatementOfTruth === Checkbox.Checked;
+  (data.applicant2StatementOfTruth === Checkbox.Checked || data.aosStatementOfTruth === Checkbox.Checked);
```
lara-mpt, line 214: there's a separate sequence for respondents, so I don't think this change is needed

src/main/steps/applicant1/hub-page/right-column/content.ts
```diff
@@ -67,6 +67,7 @@ export const generateContent: TranslationFn = content => {
   const aosSubmitted =
     !content.isJointApplication &&
     (content.userCase.applicant2StatementOfTruth ||
+      content.userCase.aosStatementOfTruth ||
```
lara-mpt, line 70: same as below, because we're checking that it's a sole application we don't need to check for applicant2StatementOfTruth

danielbe11: See NFDIV-2297

PR #2112, opened 2022-05-13
NFDIV-2221 Represented Respondent NoP posted to Respondent Address

aramchurn21:
The represented respondent Notice of Proceeding letters should show the solicitor address.

https://tools.hmcts.net/jira/browse/NFDIV-2221

- [x] tests have been updated / new tests has been added (if needed)
- [x] README and other documentation has been updated / added (if needed)


src/main/app/case/to-api-format.ts
```diff
@@ -216,6 +217,8 @@ const fields: ToApiConverters = {
   }),
   applicant1IsApplicant2Represented: data => ({
     applicant1IsApplicant2Represented: data.applicant1IsApplicant2Represented,
+    applicant2SolicitorRepresented:
+      data.applicant1IsApplicant2Represented === Applicant2Represented.YES ? YesOrNo.YES : YesOrNo.NO,
```
lara-mpt, line 221: I don't think these changes are needed anymore after one of my recent commits - https://github.com/hmcts/nfdiv-frontend/commit/c523c0309a8424c59f28ce528ae8fce652964d5f

src/main/app/case/to-api-format.ts
```diff
@@ -87,6 +87,7 @@ const fields: ToApiConverters = {
     ...(data.applicationType === ApplicationType.JOINT_APPLICATION
       ? setUnreachableAnswersToNull([
           'applicant1IsApplicant2Represented',
+          'applicant2SolicitorRepresented',
```
lara-mpt, line 90: I missed this in my changes though, so should keep this in

src/main/app/case/to-api-format.ts
aramchurn21: Ah ok, I'll checkout master as your commits are merged, and add the 'applicant2SolicitorRepresented' line to to-api-format.ts file, then I'll do a test, just to check the Notice of proceeding letters has the solicitor's address if the respondent is represented.

lara-mpt: applicant2SolicitorRepresented shouldn't need to be added to the to-api-format since it's already in case.ts
We only add to the to-api-format if the format of the field on the FE is different to the format of it in the case api

aramchurn21: Hi Lara, ok I see, thanks for the explanation. I also did a test using the latest master build and the Notice of Proceeding letters does contain the Solicitor address. Should I close this PR? I can add a comment to the ticket and assign it to Nicola.

src/main/app/case/to-api-format.ts
lara-mpt: @aramchurn21 can you update the pr to keep this line and the relevant test in please - it's something I missed in my one

aramchurn21: Ok thanks will do.

lara-mpt (commented):
@aramchurn21 looks good, just if you could add to the test for changing the application type - https://github.com/hmcts/nfdiv-frontend/blob/master/src/main/app/case/to-api-format.test.ts#L367

So add the applicant2SolicitorRepresented: null, check to make sure we're setting that to null correctly when changing to a join application

src/main/app/case/to-api-format.ts
```diff
@@ -225,6 +226,7 @@ const fields: ToApiConverters = {
         ])
       : {}),
   }),
+
```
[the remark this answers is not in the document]
aramchurn21: Sorry, yeah just removed that extra line.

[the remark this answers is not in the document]
aramchurn21: Thanks Lara, I added the applicant2SolicitorRepresented: null to to-api-format.test.

PR #2117, opened 2022-05-16
Nfdiv-2169 & 2160 - Updates to respondent solicitor fields

lara-mpt:
### Change description ###

updates to respondent solicitor fields

https://tools.hmcts.net/jira/browse/NFDIV-2160
https://tools.hmcts.net/jira/browse/NFDIV-2169


src/main/app/case/to-api-format.ts
```diff
@@ -325,6 +325,8 @@ const languagePreferenceYesNoOrNull = (value: LanguagePreference | undefined) =>
   return value === LanguagePreference.Welsh ? YesOrNo.YES : YesOrNo.NO;
 };
 
+const addressConverter = (address: (string | undefined)[]) => (address.some(Boolean) ? address.join('\n') : '');
+
```
benouaer, line 329: why do we need to do this, we can just do a for-loop in the template to display it correctly

lara-mpt: this is the to-api-format so it's not about how we display the field it's how it's sent to the backend. 

I've updated it so it just sends an empty string as oppose to "\n\n\n\n\n" which is how it was being sent before if it was empty, and therefore incorrectly showing up in xui (as per the ticket)

benouaer: so is the backend expecting it in that format, with the newlines?

benouaer (approved):

src/main/app/case/to-api-format.ts
lara-mpt: it is yeah if it's not completely empty, if it's empty then it should be a blank string

PR #2122, opened 2022-05-17
Nfdiv-2255 Updates to review your joint app page

lara-mpt:
### Change description ###
Updates to review your joint app page

https://tools.hmcts.net/jira/browse/NFDIV-2255

### DIFF
--- a/src/main/steps/applicant1/review-your-joint-application/template.njk
+++ b/src/main/steps/applicant1/review-your-joint-application/template.njk
@@ -57,11 +57,7 @@
 
   <h2 class="govuk-heading-m">{{ heading6 }}</h2>
   <p class="govuk-body">{{ line6 }}</p>
-  <ul class="govuk-list govuk-list--bullet">
-    {% for connectionBullet in connectionBulletPoints %}
-      <li>{{ connectionBullet }}</li>
-    {% endfor %}
-  </ul>
+  <p class='govuk-body'>{{ connectionBulletPoints | safe }}</p>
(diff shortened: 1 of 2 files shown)


src/main/steps/applicant1/review-your-joint-application/template.njk
```diff
@@ -57,11 +57,7 @@
 
   <h2 class="govuk-heading-m">{{ heading6 }}</h2>
   <p class="govuk-body">{{ line6 }}</p>
-  <ul class="govuk-list govuk-list--bullet">
-    {% for connectionBullet in connectionBulletPoints %}
-      <li>{{ connectionBullet }}</li>
-    {% endfor %}
-  </ul>
```
benouaer, line 64: is this not displaying the correct text?

lara-mpt: it's already displayed as a list, so it doesn't need to be turned into one. If you look at the ticket you can see a screenshot of what it looked like before.

benouaer: is that because we're making it a list in the code rather than the template?

benouaer (approved):

src/main/steps/applicant1/review-your-joint-application/template.njk
lara-mpt: yeah it was working on another page, but wasn't updated here

PR #2157, opened 2022-05-27
NFDIV-1909 - Update IrretrievablyBroken question API mapping

amoriartyMOJ:
### Change description ###

As the field for marriageIrretrievablyBroken has changed on the back end
from a YesOrNo field to a Set, this PR will update the mapping of the field.

Update definition.ts which was generated by the new API version.

Add new mapping to and from API for applicant1ScreenHasUnionBroken.

Fix issues found due to update in definition.ts file.

Update unit tests.

https://tools.hmcts.net/jira/browse/NFDIV-1909

- [X] tests have been updated / new tests has been added (if needed)
- [X] README and other documentation has been updated / added (if needed)


src/test/steps/common.ts
```diff
@@ -195,10 +194,6 @@ When('I enter my valid case reference and valid access code', async () => {
   iClick('Continue');
 });
 
-When('a case worker issues the application', async () => {
-  await triggerAnEvent(CASEWORKER_ISSUE_APPLICATION, { ceremonyPlace: 'Somewhere' });
-});
-
```
lara-mpt, line 201: Why remove this? It's used in a couple of places still

src/main/steps/applicant1/irretrievable-breakdown/post.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    formData.divorceOrDissolution = req.session.userCase.divorceOrDissolution;
+
+    if (req.body.saveAndSignOut || req.body.saveBeforeSessionTimeout) {
+      await this.saveAndSignOut(req, res, formData);
+    } else {
+      await this.saveAndContinue(req, res, form, formData);
+    }
+  }
+}
```
lara-mpt, line 24: I'm not sure this needs a custom post controller, the only difference is this sets the divorceOrDissolution field in the formData, but I can't see why we'd need to do that?

amoriartyMOJ: discussed and agreed this is the right approach as without this the divorceOrDissolution field is undefined and this causes the conversion to API to fail as it's required in order to convert correctly. Marking as resolved.

src/test/steps/common.ts
amoriartyMOJ: After chatting yesterday, this has been manually added back to get the PR working. We will be investigating why the event was removed in the first place but this PR can still progress.

lara-mpt (approved):

src/main/app/case/from-api-format.ts
```diff
@@ -19,6 +28,32 @@ const checkboxConverter = (value: string | undefined) => {
   return value === YesOrNo.YES ? Checkbox.Checked : Checkbox.Unchecked;
 };
 
+const marriageBrokenConverter = data => {
```
benouaer, line 31: do we need two converters here? could we not combine them like you do in the to-api, given that it's either DivorceOrDissolution.DIVORCE or DivorceOrDissolution.DISSOLUTION

amoriartyMOJ: From my understanding as there are 2 API fields, each one will call the same conversion which will end up with the same function being called twice if it's combined, which could change the value depending on the order it's called?

benouaer: I think the values would still match, no?

amoriartyMOJ:
Closing until requested.

PR #2232, opened 2022-06-16
Hub page clean up

lara-mpt:
### Change description ###

Create hub page template choosers to clean up hub page, and add tests


danielbe11 (approved):

src/main/steps/applicant1/hub-page/sole/soleTemplateChooser.ts
```diff
@@ -0,0 +1,54 @@
+import { State } from '../../../../app/case/definition';
+import { StateSequence } from '../../../state-sequence';
+
+export const getSoleHubTemplate = (
+  displayState: StateSequence,
+  isServiceApplicationGranted: boolean,
+  isSuccessfullyServedByBailiff: boolean,
+  isAlternativeService: boolean
+): string => {
```
benouaer, line 9: can we move

src/main/steps/applicant1/hub-page/sole/hub.njk
```diff
@@ -22,48 +22,5 @@
...
-{% elif displayState.state() == 'AwaitingAos' and isServiceApplicationGranted and not isSuccessfullyServedByBailiff %}
-  {% include "./latest-update-content/bailiff-service-unsuccessful.njk" %}
-
-{% elif displayState.state() == 'AwaitingAos' and isAlternativeService and not isServiceApplicationGranted %}
-  {% include "./latest-update-content/service-application-rejected.njk" %}
-
-{% else %}
-  {% include "./latest-update-content/aos-awaiting-or-drafted.njk" %}
-{% endif %}
+{% set theLatestUpdateTemplate = getSoleHubTemplate(displayState, isServiceApplicationGranted, isSuccessfullyServedByBailiff, isAlternativeService) %}
```
benouaer, line 25: why not work this out in the content file and pass theLatestUpdateTemplate variable into the template

src/main/steps/applicant1/hub-page/sole/soleTemplateChooser.ts
```diff
@@ -0,0 +1,54 @@
+import { State } from '../../../../app/case/definition';
```
benouaer, line 1: selector :)

src/main/steps/applicant1/hub-page/sole/soleTemplateChooser.ts
benouaer, line 9: why not just pass the userCase to the method and work out the different booleans in this function, since they're only used in this function

lara-mpt: I checked and isServiceApplicationGranted and alternativeServiceType (which sets the isAlternativeService) are both used elsewhere, so it's better to set them in the sole content file

benouaer (approved):

PR #2303, opened 2022-07-04
Fix access code post controller routes

lara-mpt:
### Change description ###

Fix access code post controller routes

### DIFF
--- a/src/main/routes.ts
+++ b/src/main/routes.ts
@@ -96,7 +96,7 @@ export class Routes {
       errorHandler(new Applicant2AccessCodeGetController().get)
     );
     app.post(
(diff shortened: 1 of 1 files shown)


nitinprabhu (approved):

src/main/routes.ts
```diff
@@ -96,7 +96,7 @@ export class Routes {
       errorHandler(new Applicant2AccessCodeGetController().get)
     );
     app.post(
-      [APPLICANT_2, RESPONDENT, `${APPLICANT_2}${ENTER_YOUR_ACCESS_CODE}`],
+      [`${APPLICANT_2}${ENTER_YOUR_ACCESS_CODE}`],
```
velimd, line 99: no need for it to be an array

PR #2390, opened 2022-07-28
Nfdiv 2350 unique urls for application submitted page 2

lara-mpt:
### Change description ###

unique urls for application submitted page 2

https://tools.hmcts.net/jira/browse/NFDIV-2350


src/main/modules/state-redirect/index.ts
```diff
@@ -40,10 +42,22 @@ export class StateRedirectMiddleware {
...
+            return res.redirect(APPLICATION_SUBMITTED);
+          }
+          if (
+            req.session.userCase.applicationType === ApplicationType.JOINT_APPLICATION &&
+            ![JOINT_APPLICATION_SUBMITTED, APPLICANT_2 + JOINT_APPLICATION_SUBMITTED].includes(req.path)
+          ) {
+            return req.session.isApplicant2
+              ? res.redirect(APPLICANT_2 + JOINT_APPLICATION_SUBMITTED)
+              : res.redirect(JOINT_APPLICATION_SUBMITTED);
+          }
```
benouaer, line 60: not sure if setting the url and doing the redirect once would be better, maybe

src/main/steps/applicant1/payment-callback/get.ts
```diff
@@ -47,7 +48,9 @@ export default class PaymentCallbackGetController {
 
     req.session.save(() => {
       if (payments.wasLastPaymentSuccessful) {
-        return res.redirect(APPLICATION_SUBMITTED);
+        return req.session.userCase.applicationType === ApplicationType.JOINT_APPLICATION
```
benouaer, line 51: I wonder if it would be better to have a function in the urls file that takes in the applicationType and returns the correct url, instead of repeating the same check every time

src/main/steps/urls.ts
```diff
@@ -96,6 +96,7 @@ export const YOUR_COMMENTS_SENT: PageLink = '/your-comments-sent';
 export const NO_RESPONSE_YET: PageLink = '/no-response-yet';
 
 export const APPLICATION_SUBMITTED: PageLink = '/application-submitted';
```
benouaer, line 98: maybe this should be a function instead that takes in the applicationType and returns the correct url accordingly

benouaer (approved):

PR #2393, opened 2022-07-29
Date translation

velimd:
### Change description ###

Enter a description.
Screenshot 2022-07-29 at 13-03-37 Gwneud cais am ysgariad - Cyflwynwyd y cais - GOV UK

https://tools.hmcts.net/jira/browse/NFDIV-2620


src/main/modules/i18n/index.ts
```diff
@@ -1,18 +1,28 @@
...
+  static supportedLanguages = [SupportedLanguages.En, SupportedLanguages.Cy];
 
   public enableFor(app: Express): void {
     app.use((req, res, next) => {
-      if (req.method === 'GET' && req.query['lng']) {
-        const requestedLanguage = req.query['lng'] as string;
-
+      if (req.method === 'GET' && (req.query['lng'] || !req.session['lang'])) {
+        // User selected language
+        const requestedLanguage = req.query['lng'] as SupportedLanguages;
```
lara-mpt, line 19: would this not cause an issue if req.query['lng'] returned undefined, and req.session['lang'] wasn't set?

lara-mpt: I'm assuming SupportedLanguages type can be undefined, in which case it would work. But why do we need to add the check of` !req.session['lang']` ? I can't see the scenario this makes a difference for

lara-mpt: the only difference I can see this making is if req.query['lng'] is undefined and so is the req.session['lang'], but all this does would be to set the req.session['lang'] to undefined again?

lara-mpt (approved):

src/main/assets/js/upload-manager/index.ts
```diff
@@ -21,7 +22,7 @@ const initUploadManager = (): void => {
   const language = document.documentElement.lang;
   location.hash = '';
 
-  const chooseFilePhoto = language === 'cy' ? 'Dewiswch ffeil' : 'Choose a file';
+  const chooseFilePhoto = language === SupportedLanguages.Cy ? 'Dewiswch ffeil' : 'Choose a file';
```
benouaer, line 25: should this be moved to the commonContent file?

benouaer (approved):

PR #2405, opened 2022-08-03
Add logging for old divorce redirect

danielbe11:
### Change description ###

- Looking into access code issue that involves redirect to old divorce


src/main/modules/oidc/index.ts
```diff
@@ -116,6 +119,7 @@ export class OidcMiddleware {
     } catch (e) {
       if (e instanceof InProgressDivorceCase) {
         const token = encodeURIComponent(req.session.user.accessToken);
+        logger.info(`Case ID ${req.session.userCase.id} being redirected to old divorce`);
```
lara-mpt, line 122: they might not have an existing NFD case at this point, so need to use 
req.session.userCase?.id

nitinprabhu (approved):

PR #2406, opened 2022-08-03
nfdiv-2585 Fix old divorce redirect

lara-mpt:
### Change description ###

Fix old divorce redirect.
Update to only send them to old divorce if they haven’t got an existing NFD case AND they haven’t been invited as a respondent/applicant2 to a NFD case

https://tools.hmcts.net/jira/browse/NFDIV-2585

### DIFF
--- a/src/main/app/case/case-api.ts
+++ b/src/main/app/case/case-api.ts
@@ -9,10 +9,6 @@ import { CASE_TYPE, CITIZEN_ADD_PAYMENT, DivorceOrDissolution, ListValue, Paymen
 import { fromApiFormat } from './from-api-format';
 import { toApiFormat } from './to-api-format';
 
-export class InProgressDivorceCase implements Error {
-  constructor(public readonly message: string, public readonly name = 'DivCase') {}
-}
-
 export class CaseApi {
   readonly maxRetries: number = 3;
 
--- a/src/main/app/case/case-api.ts
+++ b/src/main/app/case/case-api.ts
@@ -1,4 +1,3 @@
-import config from 'config';
 import { LoggerInstance } from 'winston';
 
 import { getSystemUser } from '../auth/user/oidc';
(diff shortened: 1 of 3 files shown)


nitinprabhu (approved):

src/main/modules/oidc/index.ts
```diff
@@ -79,6 +79,16 @@ export class OidcMiddleware {
 
       const existingUserCase = await req.locals.api.getExistingUserCase(res.locals.serviceType);
 
+      if (
+        !existingUserCase &&
+        !newInviteUserCase &&
+        config.get('services.case.checkDivCases') &&
+        (await req.locals.api.hasInProgressDivorceCase())
```
benouaer, line 86: we should remove the hasInProgressDivorceCase check from the getExistingUserCase function, since we're going to do it separately
also, given that we only care about existing divorce cases when there is no existing nfd case or invite case, we can add it as another else if clause in the findExistingAndNewUserCases function, and set the redirect url there

src/main/app/case/case-api.ts
```diff
@@ -38,9 +37,6 @@ export class CaseApi {
   }
 
   public async getExistingUserCase(serviceType: string): Promise<CaseWithId | false> {
-    if (config.get('services.case.checkDivCases') && (await this.hasInProgressDivorceCase())) {
-      throw new InProgressDivorceCase('User has in progress divorce case');
```
benouaer, line 42: I guess we don't need the InProgressDivorceCase exception class anymore, since we won't throw the exception

src/main/modules/oidc/index.ts
lara-mpt: we can, I added this to the top instead because I thought that if/else block is getting a bit ugly and this might be clearer (and clearer for the issue at the moment). If we have time at some point I'd like to try cleaning that block up a bit. I can move it now if you prefer it there though

benouaer: it's not a big deal, just that the hasInProgressDivorceCase is only relevant if there are no existing or invite user cases. so we can save ourselves having to do the check by putting it towards the end

lara-mpt: ah I see, that makes sense. Will update that - thanks

benouaer (approved):

PR #2418, opened 2022-08-08
NFDIV-2553 Submit AOS bug (v3)

danielbe11:
### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-


src/main/assets/js/confirm-read-petition-submit.ts
```diff
@@ -0,0 +1,9 @@
+import { getById } from './selectors';
+
+const confirmReadPetitionField = getById('confirmReadPetitionId') as HTMLFormElement | null;
+
+if (confirmReadPetitionField && confirmReadPetitionField.getAttribute('disabled')) {
+  (getById('main-form') as HTMLFormElement).onsubmit = () => {
+    confirmReadPetitionField.removeAttribute('disabled');
+  };
```
lara-mpt, line 8: shouldn't the default be 'enabled' and then onsubmit you disable it?

danielbe11: All this bit of code is doing is removing the 'disabled' attribute upon submit whenever disabled=true. This is important because if you submit a disabled field then the value isn't actually sent in the POST request. 
Choosing whether to disable the field is done elsewhere.

danielbe11: elsewhere = https://github.com/hmcts/nfdiv-frontend/pull/2418/files#diff-ff1ec5ce83c678241cfdb14a892a1ce763502e97c06dc944767bfbe8d06b922bR20

lara-mpt (approved):

PR #2511, opened 2022-08-30
NFDIV-2740 - Prevent multiple cases creation

danielbe11:
### Change description ###

- Move case creation to your-details to avoid double click => double creation of cases.

https://tools.hmcts.net/jira/browse/NFDIV-2740


src/main/steps/home/get.ts
```diff
@@ -57,6 +57,10 @@ const getApplicant2FirstQuestionForm = (applicationType: ApplicationType) =>
   applicationType === ApplicationType.SOLE_APPLICATION ? respondentFirstQuestionForm : applicant2FirstQuestionForm;
 
 const applicant1RedirectPageSwitch = (userCase: Partial<CaseWithId>, isFirstQuestionComplete: boolean) => {
+  if (!userCase) {
+    return YOUR_DETAILS_URL;
+  }
```
lara-mpt, line 62: this check could go in the initial function/logic

src/main/assets/js/disabled-upon-submit.ts
```diff
@@ -0,0 +1,12 @@
+import { getById } from './selectors';
+
+const disableUponSubmit = (): void => {
+  const submitButton = getById('main-form-submit') as HTMLFormElement | null;
+  const form = getById('main-form') as HTMLFormElement | null;
+  const currentUrl = document.location.pathname;
+  if (form && submitButton && currentUrl === '/your-details') {
```
lara-mpt, line 7: better to import the YOUR_DETAILS url than hardcode it

src/main/steps/applicant1/your-details/post.ts
```diff
@@ -9,6 +9,10 @@ import { setJurisdictionFieldsAsNull } from '../../../app/jurisdiction/jurisdict
 @autobind
 export default class YourDetailsPostController extends PostController<AnyObject> {
   public async post(req: AppRequest<AnyObject>, res: Response): Promise<void> {
+    if (!req.session.userCase) {
+      req.session.userCase = await req.locals.api.createCase(res.locals.serviceType, req.session.user);
+    }
```
lara-mpt, line 14: don't some tests need to be added for this?

lara-mpt (approved):

src/main/modules/oidc/index.ts
```diff
@@ -108,15 +108,15 @@ export class OidcMiddleware {
             return next();
           }
         }
-        req.session.userCase =
-          req.session.userCase ||
-          existingUserCase ||
-          (await req.locals.api.createCase(res.locals.serviceType, req.session.user));
+        req.session.userCase = req.session.userCase || existingUserCase;
 
         req.session.existingCaseId = req.session.userCase.id;
```
benouaer, line 113: req.session.userCase might be null, so we need to protect against that here

src/main/assets/js/disabled-upon-submit.ts
```diff
@@ -0,0 +1,12 @@
...
+    form.addEventListener('submit', () => submitButton.setAttribute('disabled', 'true'));
```
benouaer, line 8: can we make this a generic thing rather than just for this particular page
we're seeing a similar issue with the applyForConditionalOrder button when it's clicked twice

danielbe11: Done

PR #2518, opened 2022-08-31
NFDIV-2785 - Fix timeout page issues

danielbe11:
### Change description ###

Two issues:

- No content showing upon timeout
- After 20 mins on the timeout page, it will sign out. I think it would be better to give the timeout page 12 hours (like webchat) before timeout.

https://tools.hmcts.net/jira/browse/NFDIV-2785


lara-mpt (approved):

src/main/assets/js/session-timeout.ts
```diff
@@ -41,6 +40,17 @@ const pingUserActive = throttle(
...
 );
 
+const getSessionTimeoutInterval = (): number => {
+  const timeoutParam = new URL(location.href).searchParams.get('timeout');
+  if (timeoutParam && !isNaN(parseInt(timeoutParam))) {
+    return parseInt(timeoutParam);
+  }
+
+  return [WEBCHAT_URL, TIMED_OUT_URL].includes(window.location.pathname as PageLink)
+    ? 12 * 60 * 60 * 1000 // 12 hours
```
benouaer, line 50: why does the timedOut url result in a 12hr timeout?

src/main/assets/js/session-timeout.ts
```diff
@@ -41,6 +40,17 @@ const pingUserActive = throttle(
   { trailing: false }
 );
 
+const getSessionTimeoutInterval = (): number => {
+  const timeoutParam = new URL(location.href).searchParams.get('timeout');
```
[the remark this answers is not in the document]
danielbe11: This is only supposed to just be for testing purposes, no citizen users should be setting this timeout, but I'll still add something

src/main/assets/js/session-timeout.ts
danielbe11: Spoke to Michael about this - he wants timeout page to show indefinitely so the user can see that they've been signed out for privacy reasons and their data has been saved. The way it works atm is that if the user was to leave their PC for over 40mins, they would then see IDAM page with no explanation on why they have logged out. 
It's safe to leave them on timeout page indefinitely because their session has already been destroyed.

benouaer (approved):

PR #2568, opened 2022-09-21
NFDiv-2874 Add Submit Property to Common.Content

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-2874


lara-mpt (approved):

velimd (approved):
there one more place i believe its used which is in check your conditional order answers. might be best to remove it there as well.

lara-mpt (approved):

src/test/functional/features/joint-conditional-order.feature
```diff
@@ -35,7 +35,7 @@ Feature: Joint conditional order
     Then the page URL should be "/check-your-conditional-order-answers"
 
     Given I select "I believe that the facts stated in this application are true"
-    When I click "Continue"
+    When I click "Submit"
```
benouaer, line 38: why has the submit button text changed?

BWeeks-JusticeGovUK: The content file had defined a submit property (with values equivalent to those that I have set in the common content) which I believe was intended for the button text - however it had been improperly applied.  Therefore I corrected the content to pull from common, and applied the value correctly to the button.  This in turn required updating the tests.

benouaer (approved):

PR #2602, opened 2022-10-06
NFDIV-3048 - Fix case priority 

danielbe11:
### Change description ###

- Priority case selection is not working correctly.

https://tools.hmcts.net/jira/browse/NFDIV-3048


src/main/steps/state-sequence.ts
```diff
@@ -38,52 +38,54 @@ export class StateSequence {
...
-    State.FinalOrderRequested,
-    State.FinalOrderPending,
-    State.FinalOrderComplete,
-    State.Rejected,
-    State.Withdrawn,
-    State.NewPaperCase,
-    State.OfflineDocumentReceived,
-    State.BulkCaseReject,
-  ]).at(userCase.state as State);
+  return new StateSequence(chronologicalStateSequence).at(userCase.state as State);
```
lara-mpt, line 41: something that could be done as a separate piece of work but StateSequence.at() shouldn't be returning a new StateSequence, as it's being called by the StateSequence itself

danielbe11: Created NFDIV-3065.

src/main/app/case/case-api.test.ts
```diff
@@ -316,6 +333,11 @@ describe('CaseApi', () => {
     const isApplicant2 = await api.isApplicant2('1234123412341234', userDetails.id);
     expect(isApplicant2).toBe(true);
   });
+
+  test('preSubmittedStatePrioritySequence should be appropriate', async () => {
```
benouaer, line 337: I think you can test this through the result of Submitted not being in the preSubmittedStatePrioritySequence, rather than exporting it just for testing

src/main/app/case/case-api.test.ts
```diff
@@ -2,7 +2,7 @@ import * as oidc from '../auth/user/oidc';
 import { UserDetails } from '../controller/AppRequest';
 import { PaymentModel } from '../payment/PaymentModel';
 
-import { CaseApi, getCaseApi } from './case-api';
+import { CaseApi, exportedForTesting, getCaseApi } from './case-api';
```
benouaer, line 5: I would be suspicious of an export just for testing

src/main/app/case/case-api.ts
```diff
@@ -98,33 +99,31 @@ export class CaseApi {
...
+      const submittedUserCase = userCases.find(userCase => !preSubmittedStatePrioritySequence.includes(userCase.state));
+      if (submittedUserCase) {
+        return convertCcdV1ResponseToCaseWithId(submittedUserCase);
       }
+      // If we have two Draft cases, this method is safe in finding the latest case because the [0] element will be the latest case.
+      // Otherwise, the highest priority pre-submitted case will be chosen (the one with the highest index):
+      return convertCcdV1ResponseToCaseWithId(this.getHighestPriorityPreSubmissionCases(userCases)[0]);
     }
+    // Else, return the only case we have (or false):
     return this.getLatestUserCase(userCases);
```
lara-mpt, line 113: You could do
`const userCase = submittedUserCase || userCases[0]`
and then only need to return the convertCcdV1ResponseToCaseWithId once
`return convertCcdV1ResponseToCaseWithId(userCase); `

src/main/steps/state-sequence.ts
```diff
@@ -38,53 +38,55 @@ export class StateSequence {
...
 };
+
+export const chronologicalStateSequence: State[] = [
```
benouaer, line 44: orderedStateSequence?

src/main/app/case/case-api.ts
```diff
@@ -98,33 +99,31 @@ export class CaseApi {
...
   }
 
+  private getHighestPriorityPreSubmissionCases(userCases: CcdV1Response[]): CcdV1Response[] {
+    const stateIndexArr = userCases.map(userCase => preSubmittedStatePrioritySequence.indexOf(userCase.state));
+    const highestPriorityStateIndex: number = Math.max(...stateIndexArr);
```
benouaer, line 118: maybe we could add this as a function in the state-sequence class, as we already have before and after methods there.

danielbe11:
Closed and moved to a cleaner PR https://github.com/hmcts/nfdiv-frontend/pull/2607

PR #2619, opened 2022-10-18
Cache test IDAM tokens

danielbe11:
### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-


nitinprabhu (approved):

src/test/steps/IdamUserManager.ts
```diff
@@ -89,6 +94,17 @@ export class IdamUserManager {
     this.users = new Set([firstUser]);
   }
 
+  static async getAccessTokenFromIdam(username: string, password: string): Promise<AxiosResponse<OidcResponse>> {
+    const id: string = sysConfig.get('services.idam.clientID');
+    const secret = sysConfig.get('services.idam.clientSecret');
+    const tokenUrl: string = sysConfig.get('services.idam.tokenURL');
+
+    const headers = { Accept: 'application/json', 'Content-Type': 'application/x-www-form-urlencoded' };
+    const data = `grant_type=password&username=${username}&password=${password}&client_id=${id}&client_secret=${secret}&scope=openid%20profile%20roles%20openid%20roles%20profile`;
```
benouaer, line 103: I know this is how it was setup before, but can we use URLSearchParams instead of appending everything to the url 
it should make it much more readable (unless it doesn't actually work)

src/main/app/auth/user/oidc.ts
```diff
@@ -39,17 +43,24 @@ export const getUserDetails = async (
...
+  const isNotProd: boolean = process.env.NODE_ENV !== 'production';
+  if (idamTokenCache.get(systemUsername) && isNotProd) {
+    logger.info('Fetching systemUsername from cache...');
+    response = idamTokenCache.get(systemUsername);
+  } else if (isNotProd) {
+    logger.info('Generating access token for systemUsername...');
+    response = await IdamUserManager.getAccessTokenFromIdam(systemUsername, systemPassword);
+    idamTokenCache.set(systemUsername, {
+      data: { id_token: response.data.id_token, access_token: response.data.access_token },
+    });
```
benouaer, line 59: do we know how it takes for the idam token to expire, I assume the cache would need to use the same expiry duration. also, do we need a dependency for this, could we not just store token in-session ourselves?

charts/nfdiv-frontend/values.aat.template.yaml
```diff
@@ -8,6 +8,7 @@ nodejs:
     primaryIngressHost: ${SERVICE_FQDN}
   environment:
     OAUTH_CLIENT_REDIRECT: https://${SERVICE_FQDN}/oauth2/callback
+    IDAM_TOKEN_CACHE: true
```
benouaer, line 11: do you need to set this, since you're doing it in flux?

danielbe11: @nitinprabhu said this affects staging env, which is distinct from AAT (which is configured using flux)

benouaer (approved):

PR #2626, opened 2022-10-21
NFDIV-3090 - Prevent AOS being updated after submit

danielbe11:
### Change description ###

- It is possible to go back to AOS URLs on the FE to change answers. This shouldn't be possible.

https://tools.hmcts.net/jira/browse/NFDIV-3090


src/main/steps/index.utils.ts
```diff
@@ -0,0 +1,5 @@
+import { CaseWithId } from '../app/case/case';
+
+export const hasSubmittedAos = (userCase: CaseWithId): boolean => {
```
benouaer, line 3: do we need a whole new utils file for just a boolean? it's only used in one place for now right? so we could put this function in the index file

src/main/steps/respondentSequence.ts
```diff
@@ -89,7 +89,6 @@ const sequence: Step[] = [
 
 export const respondentSequence = ((): Step[] => {
   return sequence.map(step => ({
-    ...step,
```
benouaer, line 92: do we not need the other step properties?

benouaer (approved):

src/main/steps/index.utils.ts
danielbe11: Moved it to index

src/main/steps/respondentSequence.ts
danielbe11: This ...step was added when our steps looked like this:
  {
    url: DETAILS_OTHER_PROCEEDINGS,
    showInSection: Sections.OtherCourtCases,
    getNextStep: () => HOW_THE_COURTS_WILL_CONTACT_YOU,
  },
but now we don't have showInSection we can remove it.

PR #2643, opened 2022-10-26
NFDiv-3127 Refactor template selectors

BWeeks-JusticeGovUK:
Refactor Sole and Joint template selectors to utilise enums instead of strings.


benouaer (approved):

danielbe11 (commented):
LGTM, nice work

PR #2698, opened 2022-11-17
NFDiv-2697 fo switch to sole finalising divorce as sole

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-2697


src/main/steps/common/switch.to.sole.content.utils.test.ts
```diff
@@ -0,0 +1,130 @@
+import dayjs from 'dayjs';
```
danielbe11, line 1: I think the name should be src/main/steps/common/switch-to-sole-content.utils.test.ts. The purpose of the dots is to separate the purpose of the file i.e. a utils file or a tests file.

BWeeks-JusticeGovUK: Thanks - will resolve in the next commit.

src/main/steps/applicant1/finalising-your-application/post.ts
```diff
@@ -19,7 +30,39 @@ export default class FinalisingYourApplicationPostController extends PostControl
     return super.save(req, formData, eventName);
   }
 
+  protected getNextUrl(req: AppRequest): string {
+    const hasApplicant2SwitchedToSoleFo =
+      req.session.isApplicant2 &&
+      req.originalUrl === APPLICANT_2 + FINALISING_YOUR_APPLICATION &&
+      req.session.userCase.finalOrderSwitchedToSole === YesOrNo.YES &&
+      req.session.userCase.state === State.FinalOrderRequested &&
```
danielbe11, line 38: Need to be careful when the case is Welsh. Should be looking at Welsh previous state

danielbe11: Could just leave this line out if thats gonna complicate this

BWeeks-JusticeGovUK: Removed the state check - the other logical tests should be sufficient.

src/main/steps/common/switch-to-sole-content.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const hasApplicantDeclaredIntentionToSwitchToSoleFinalOrder = (
+  userCase: Partial<CaseWithId>,
+  isApplicant2: boolean
+): boolean => {
+  return (
+    hasApplicantAppliedForFinalOrderFirst(userCase, isApplicant2) &&
+    doesApplicantIntendToSwitchToSoleFinalOrder(userCase, isApplicant2) &&
+    dayjs().isAfter(dateApplicantDeclaredIntentionToSwitchToSoleFinalOrder(userCase, isApplicant2))
```
danielbe11, line 40: Do we need both?

danielbe11: Adding too many conditions that check the same sort of thing might be a bit confusing to people in the future

src/main/steps/common/switch-to-sole-content.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+  );
+};
+
+const applicantSwitchToSoleFinalOrderIntentionNotificationPeriodExpiryDate = (
```
danielbe11, line 44:
```suggestion
const getApplicantSwitchToSoleFinalOrderIntentionNotificationPeriodExpiryDate = (
```

src/main/steps/common/switch-to-sole-content.utils.ts
BWeeks-JusticeGovUK: It was setting up a check elsewhere which should have a start and end date.  However this is only required in one other location and not each time this method is called.  I have refactored this.

danielbe11: Can we move it to isWithinSwitchToSoleFinalOrderIntentionNotificationPeriod method

danielbe11 (approved):

src/main/steps/applicant1/finalising-your-application/post.ts
```diff
@@ -19,7 +29,37 @@ export default class FinalisingYourApplicationPostController extends PostControl
...
 
+  protected getNextUrl(req: AppRequest): string {
+    const hasApplicant2SwitchedToSoleFo =
+      req.session.isApplicant2 &&
+      req.originalUrl === APPLICANT_2 + FINALISING_YOUR_APPLICATION &&
+      req.session.userCase.finalOrderSwitchedToSole === YesOrNo.YES &&
+      this.formData.doesApplicant2WantToApplyForFinalOrder;
+
+    if (hasApplicant2SwitchedToSoleFo) {
+      req.session.isApplicant2 = false;
```
benouaer, line 40: not sure this is the right place to be changing this flag, because it has wider implications apart from the nextUrl
I think it should probably be updated when the event to switch to sole is submitted
there might be other things in the session that also need to be updated as a result of that

src/main/steps/applicant1/finalising-your-application/post.ts
```diff
@@ -19,7 +29,37 @@ export default class FinalisingYourApplicationPostController extends PostControl
...
+    return super.getNextUrl(req);
+  }
+
+  protected async saveAndContinue(
+    req: AppRequest<AnyObject>,
+    res: Response,
+    form: Form,
+    formData: Partial<Case>
+  ): Promise<void> {
+    this.formData = formData;
```
benouaer, line 53: not sure what this is doing

src/main/steps/applicant1/hub-page/joint/content.ts
```diff
@@ -306,54 +308,41 @@ export const generateContent: TranslationFn = content => {
...
-    userCase,
+  const theLatestUpdateTemplate = getJointHubTemplate(displayState, userCase, {
     hasApplicantAppliedForConditionalOrder,
-    withinSwitchToSoleFinalOrderIntentionNotificationWindow
-  );
+    isWithinSwitchToSoleFoIntentionNotificationPeriod:
+      switchToSoleFinalOrderStatus.isWithinSwitchToSoleFoIntentionNotificationPeriod,
+    hasSwitchToSoleFoIntentionNotificationPeriodExpired:
+      switchToSoleFinalOrderStatus.hasSwitchToSoleFoIntentionNotificationPeriodExpired,
+  });
```
benouaer, line 346: why not just pass switchToSoleFinalOrderStatus, this seems to be recreating the same object

benouaer (approved):

src/main/steps/applicant1/hub-page/joint/content.ts
BWeeks-JusticeGovUK: I was working on the principle of only passing the properties that are required.

src/main/steps/applicant1/finalising-your-application/post.ts
BWeeks-JusticeGovUK: Grabbing the formData and assigning it to the instance of the class so that it's available for the local instance of another method.

src/main/steps/applicant1/finalising-your-application/post.ts
BWeeks-JusticeGovUK: We shouldn't change it on submit - we should only change it if the switch to sole event was successful.  At that point the applicant data has been entirely swapped, and we should be switching the flag.  I'm not sure if anything else should also be updated - it's possible, but at this point the end user journey is complete and they will only have the option to logout.  I believe the risk is negligible.

BWeeks-JusticeGovUK: I realised that an error count check was missing from hasApplicant2SwitchToSoleFo - this has been corrected.

src/main/steps/applicant1/finalising-your-application/post.ts
BWeeks-JusticeGovUK: Refactored and removed the requirement for the formData.

PR #2704, opened 2022-11-21
NFDIV-3153: corrected marriage date validation

sanjayparekh-hmcts:
### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-313

### DIFF
--- a/src/main/app/form/validation.ts
+++ b/src/main/app/form/validation.ts
@@ -73,16 +73,17 @@ export const isFutureDate: DateValidator = date => {
   }
 };
 
-export const isLessThanAYear: DateValidator = date => {
+export const isMoreThanAYearAgo: DateValidator = date => {
   if (!date) {
     return;
   }
 
   const enteredDate = new Date(+date.year, +date.month - 1, +date.day);
-  const oneYearAgo = new Date();
+  const today = new Date();
+  const oneYearAgo = new Date(today.getFullYear(), today.getMonth(), today.getDate());
   oneYearAgo.setFullYear(oneYearAgo.getFullYear() - 1);
-  if (!(enteredDate < oneYearAgo)) {
-    return 'lessThanAYear';
+  if (enteredDate < oneYearAgo) {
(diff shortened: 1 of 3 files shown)


src/main/app/form/validation.ts
```diff
@@ -73,16 +73,17 @@ export const isFutureDate: DateValidator = date => {
...
 
   const enteredDate = new Date(+date.year, +date.month - 1, +date.day);
-  const oneYearAgo = new Date();
+  const today = new Date();
+  const oneYearAgo = new Date(today.getFullYear(), today.getMonth(), today.getDate());
   oneYearAgo.setFullYear(oneYearAgo.getFullYear() - 1);
-  if (!(enteredDate < oneYearAgo)) {
-    return 'lessThanAYear';
+  if (enteredDate >= oneYearAgo) {
+    return 'lessThanAYearAgoInc';
```
benouaer, line 86: I know this is how it was before, but to me it looks like the validators should only return a string when there is an error, not when the validation is successful. because otherwise they should return a boolean, but the string return is to tell us what type of error it is

sanjayparekh-hmcts: Yes - I tried it the other way around, and it caused a lot of problems - see my comments on the JIRA

src/main/app/form/validation.ts
```diff
@@ -73,16 +73,17 @@ export const isFutureDate: DateValidator = date => {
...
-export const isLessThanAYear: DateValidator = date => {
+export const isLessThanAYearAgoInc: DateValidator = date => {
   if (!date) {
     return;
   }
 
   const enteredDate = new Date(+date.year, +date.month - 1, +date.day);
-  const oneYearAgo = new Date();
+  const today = new Date();
+  const oneYearAgo = new Date(today.getFullYear(), today.getMonth(), today.getDate());
```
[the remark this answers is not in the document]
sanjayparekh-hmcts: ok will do

benouaer (approved):

PR #3199, opened 2023-09-13
remove final order overdue state

linusnorton:
### Change description ###

Remove the FinalOrderOverdue state and rely on isFinalOrderOverdue instead

### DIFF
--- a/src/main/steps/applicant1/hub-page/joint/jointTemplateSelector.ts
+++ b/src/main/steps/applicant1/hub-page/joint/jointTemplateSelector.ts
@@ -40,7 +40,6 @@ export const getJointHubTemplate = (
     case State.AwaitingAdminClarification:
     case State.AwaitingLegalAdvisorReferral:
       return HubTemplate.AwaitingLegalAdvisorReferral;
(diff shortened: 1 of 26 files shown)


src/main/steps/applicant1/hub-page/joint/jointTemplateSelector.ts
```diff
@@ -40,7 +40,6 @@ export const getJointHubTemplate = (
     case State.AwaitingAdminClarification:
     case State.AwaitingLegalAdvisorReferral:
       return HubTemplate.AwaitingLegalAdvisorReferral;
-    case State.FinalOrderOverdue:
```
velimd, line 43: i wonder if this will break existing functionality. i.e. app1 applies -> overdue, app1 doesnt seem this page now as its in awaitingjointfinalorder.

linusnorton: I think that's correct. If app1 has done what they need to, then they shouldn't get any more hassle about it being late

velimd (approved):

PR #3860, opened 2024-09-03
[NFDIV-4314] Prevent multiple service requests when payments are created

adamg-hmcts:
### Change description ###
F&P have identified that multiple service requests are still being created for some payments and we need to make changes to the payments workflow to avoid this.

**1.) Old frontend payment workflow:**
Check case data for payments > Create service request if no payment found > Attempt payment > Submit citizen-add-payment event

**Problem:** If the citizen-add-payment event fails, the payment and service request are not saved to our case data, meaning we do not have a record of the service request reference created in the F&P DB.

**2.) New frontend payment workflow:**
Check case payment groups (service requests) via GET request to F&P API > Create payment with new service request if none found, or reattempt payment using the same service request if one is found > Submit citizen-add-payment event

https://tools.hmcts.net/jira/browse/NFDIV-4314


src/main/app/controller/BasePaymentPostController.ts
```diff
@@ -105,14 +105,26 @@ export default abstract class BasePaymentPostController {
       CITIZEN_ADD_PAYMENT
     );
 
-    return payment;
+    return redirectUrl;
+  }
+
+  public async findServiceRequestReference(feeCode: string, client: PaymentClient): Promise<string | undefined> {
```
DelythJustice, line 111: I don't understand this bit.
Is it calling into f&p and only find the service request made for this case? Or is it finding a service request made with the fee code?

DelythJustice: thinking we need to be careful we don't just grab a service request but that it belongs to someone else's case

adamg-hmcts: Hi @DelythJustice, we get the payment groups for their case by calling `/cases/${caseId}/paymentgroups` in [PaymentClient.getCasePaymentGroups()](https://github.com/hmcts/nfdiv-frontend/pull/3860/files#diff-78fbef39087b4e70a8b5cb5892d7f1b0fdd5776fbbd568dbc5b746a3a422c384R31). CaseId hasn't been passed as an arg here because we can get it from the user session in payment client

DelythJustice (approved):

src/main/app/controller/BasePaymentPostController.ts
DelythJustice: ta

PR #3990, opened 2024-11-12
NFDIV-4439 Add functional tests for RFI frontend journey

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4439

### DIFF
--- a/src/test/functional/features/joint-rfi-response.feature
+++ b/src/test/functional/features/joint-rfi-response.feature
@@ -1,0 +1,219 @@
+Feature: Joint request for information
+
+  Background: Logged in for hub page
+    Given I create a new user and login
+    And I've already completed the form using the fixture "jointApplicant1CompleteCase"
+    When I go to "/"
+    And I click "Send for review"
+    Then the page URL should be "/application-sent-for-review"
+    When I enter my valid case reference and valid access code
+    Then the page should include "You need to review your joint application"
+    Given I've already completed the form using the fixture "jointApplicant2CompleteCase" for "applicant2"
+    And I go to "/applicant2/confirm-your-joint-application"
+    Given I select "I confirm that I’m applying to the court to dissolve my marriage (get a divorce)"
+    And I select "I believe that the facts stated in this application are true"
+    When I click "Submit"
+    Then the page URL should be "/applicant2/needs-to-confirm-joint-application"
+
+  @nightly
+  Scenario: [1] Applicant 1 joint rfi journey
+    Given a case worker issues a request for information to app1 on a joint case
+    When I click "Sign out"
+    And I login with applicant "2"
+    Then the page URL should be "/applicant2/hub-page"
+    And the page should include "The court has reviewed your application for divorce. We have sent an email to your wife with the information that the court needs."
+    And the page should include "The court will review the information from your wife once provided, then the application can progress."
+    When I click "Sign out"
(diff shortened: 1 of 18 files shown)


src/test/functional/features/joint-rfi-response.feature
```diff
@@ -0,0 +1,249 @@
...
+    Then the page should include "You need to review your joint application"
+    Given I've already completed the form using the fixture "jointApplicant2CompleteCase" for "applicant2"
+    And I go to "/applicant2/confirm-your-joint-application"
+    Given I select "I confirm that I’m applying to the court to dissolve my marriage (get a divorce)"
+    And I select "I believe that the facts stated in this application are true"
+    When I click "Submit"
+    Then the page URL should be "/applicant2/needs-to-confirm-joint-application"
+
+  @nightly
+  Scenario: [1] Applicant 1 joint rfi journey
```
adamg-hmcts, line 18: Do you think we should make the happy path run outside of nightly to reduce the chance of people introducing RFI bugs by only relying on normal CI?

src/test/functional/features/joint-rfi-response.feature
```diff
@@ -0,0 +1,249 @@
...
+    Given a case worker issues a request for information to app1 on a joint case
+    When I click "Sign out"
+    And I login with applicant "2"
+    Then the page URL should be "/applicant2/hub-page"
+    And the page should include "The court has reviewed your application for divorce. We have sent an email to your wife with the information that the court needs."
+    And the page should include "The court will review the information from your wife once provided, then the application can progress."
```
adamg-hmcts, line 24: It might be best to match against shorter strings of text to speed up the tests and make them less brittle. Matching against element IDs is often suggested as the best practise because they don't change as often as text, meaning less need to update the test

src/test/functional/features/joint-rfi-response.feature
```diff
@@ -0,0 +1,249 @@
...
+    When I click "Continue"
+    Then the page URL should be "/review-your-response-to-the-courts-feedback"
+    And the page should include "Review your response"
+
+    When I click "Submit"
+    Then the page URL should be "/hub-page"
+    And the page should include "You have responded to the court."
+
+  @nightly
+  Scenario: [1] Applicant 2 joint rfi journey
```
adamg-hmcts, line 45: How slow are you finding these tests are? Thinking about whether we can test some of the edge cases effectively without functional tests (testing pyramid)

src/test/functional/features/joint-rfi-response.feature
BWeeks-JusticeGovUK: Yes - I have removed the @nightly decorator from the sole applicant and joint app1 and 2 happy paths

src/test/functional/features/joint-rfi-response.feature
BWeeks-JusticeGovUK: I would need to remove the @nightly decorator and run them in the pipeline to be sure - everything is slow locally!

src/test/functional/features/joint-rfi-response.feature
adamg-hmcts: Thanks @BWeeks-JusticeGovUK, if it's easier to shorten the text that should be fine as well

adamg-hmcts (approved):

PR #4122, opened 2025-03-19
[NFDIV-4713] No Response Feature Branch

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4713


src/main/steps/applicant1/hub-page/sole/content.ts
```diff
@@ -778,6 +808,11 @@ export const generateContent: TranslationFn = content => {
   const isRespondentRepresented = userCase.applicant1IsApplicant2Represented === Applicant2Represented.YES;
   const isAosSubmitted = !isEmpty(userCase.dateAosSubmitted);
   const aosIsDrafted = userCase.aosIsDrafted === YesOrNo.YES;
+  const aosOverdueAndDrafted =
+    userCase.state === State.AosDrafted &&
+    !userCase.aosStatementOfTruth &&
+    userCase.issueDate &&
```
adamg-hmcts, line 814: Thinking we might want to use the `aosIsDrafted` variable that's based on the case data rather than checking for State.AosDrafted? The case can be shifted from AosDrafted state to AosOverdue state and it looks like that would stop the user seeing the text to let them know the Aos has been drafted?

adamg-hmcts (approved):

PR #4219, opened 2025-06-18
NFDIV-4860: Update hub page for contact details updated

FaisalMoJ:
### Change description ###

Update hub page for contact details updated

https://tools.hmcts.net/jira/browse/NFDIV-4860

### DIFF
--- a/src/main/app/case/definition.ts
+++ b/src/main/app/case/definition.ts
@@ -2493,6 +2493,7 @@ export const enum NoResponsePartnerNewEmailOrPostalAddress {
   NEW_POSTAL = 'newPostalAddress',
   NEW_EMAIL = 'newEmailAddress',
   BOTH_EMAIL_AND_POSTAL = 'newEmailAndPostalAddress',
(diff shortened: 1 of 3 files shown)


src/main/app/case/definition.ts
```diff
@@ -2493,6 +2493,7 @@ export const enum NoResponsePartnerNewEmailOrPostalAddress {
   NEW_POSTAL = 'newPostalAddress',
   NEW_EMAIL = 'newEmailAddress',
   BOTH_EMAIL_AND_POSTAL = 'newEmailAndPostalAddress',
+  CONTACT_DETAILS_UPDATED = 'contactDetailsUpdated',
```
adamg-hmcts, line 2496: Is the user able to select this new value in an input somewhere?

adamg-hmcts (approved):

src/main/app/case/definition.ts
FaisalMoJ: Thanks Adam i will test this locally but i also think there wouldn't be any problem since this doesn't exist for users in the journey

PR #4221, opened 2025-06-18
NFDIV-4763 Dispense with service 1

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4763


src/main/steps/dispenseServiceApplicationSequence.ts
```diff
@@ -1,9 +1,63 @@
...
+    url: LAST_ADDRESS_DISPENSE,
+    getNextStep: () => AWARE_PARTNER_ADDRESS_DISPENSE,
+  },
+  {
+    url: AWARE_PARTNER_ADDRESS_DISPENSE,
+    getNextStep: () => PARTNER_NEW_ADDRESS_DISPENSE,
+  },
+  {
+    url: PARTNER_NEW_ADDRESS_DISPENSE,
+    getNextStep: () => LAST_SEEN_DISPENSE,
```
adamg-hmcts, line 57: There are some page skips in Miro that it looks like we might need to add logic for here:
image

PR #4831, opened 2026-06-16
NFDIV-5618 Update Frontend Dependencies

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-5618


package.json
```diff
@@ -134,6 +135,7 @@
     "eslint-plugin-import": "2.29.1",
     "eslint-plugin-jest": "29.15.2",
     "eslint-plugin-prettier": "5.5.6",
+    "form-data": "^2.5.6",
```
adamg-hmcts: It's showing that @opentelemetry/core and form-data have been added as new project dependencies. The vulns in those look to have come in as transitive dependencies (sub-dependencies), which we can review the source of in yarn.lock and by using `yarn why`. We have four options to address those (from best to worst):

1.) Upgrading the dependency that uses the vulnerable subdependency, as the latest version of the dependency may have upgraded to newer versions of the subdependencies.
2.) Adding a resolution in package.json to force project dependencies to use a newer version of the subdependency.
3.) Remove or replace the dependency that brings in the vulnerable subdependency, if it's not actively maintained or has multiple unresolved vulnerabilities.
4.) Accept it as a known vulnerability, if it cannot be addressed and has been assessed as not introducing risk.

Adding the packages as direct dependencies isn’t ideal in this case, as it doesn’t force vulnerable versions in the dependency tree to be changed to address the CVE, and can lead to duplicated/unused packages.

BWeeks-JusticeGovUK: Yeah - I don't know why I put them into devDependencies instead of Resolutions.

I've refactored, and updated Axios so no resolution is required for form-data.  Upgrading applicationInsights prevented the build from deploying to preview, so I've left that alone and added a resolution for @opentelemetry/core.

adamg-hmcts (approved):

PR #4854, opened 2026-06-29
NFDIV-5465: Refactor review contact details

FaisalMoJ:
### Change description ###

Refactor review contact details

https://tools.hmcts.net/jira/browse/NFDIV-5465


src/main/steps/respondentSequence.ts
```diff
@@ -22,18 +23,44 @@ import {
...
+    getNextStep: data => (data.applicant2AddressPrivate === YesOrNo.NO ? ENTER_YOUR_ADDRESS : IN_REFUGE),
+  },
+  {
+    url: IN_REFUGE,
+    getNextStep: () => ENTER_YOUR_ADDRESS,
+  },
+  {
+    url: ENTER_YOUR_ADDRESS,
+    getNextStep: data =>
+      data.applicant2ConfirmContactDetails === YesOrNo.NO ? CHECK_PHONE_NUMBER : HOW_DO_YOU_WANT_TO_RESPOND,
```
adamg-hmcts: I can't see branching logic for this page in the UCD designs, do we need this logic in place?

FaisalMoJ: It looks like that Miro board might have been updated as i am not sure why this logic is there. I will update it according to Miro

adamg-hmcts (approved):

src/main/app/case/to-api-format.ts
```diff
@@ -721,6 +721,12 @@ const fields: ToApiConverters = {
         ? data.applicant1DispenseChildMaintenanceResults
         : null,
   }),
+  applicant2ConfirmContactDetails: data => ({
+    applicant2ConfirmContactDetails: data.applicant2ConfirmContactDetails,
+    ...(data.applicant2ConfirmContactDetails === YesOrNo.YES
+      ? setUnreachableAnswersToNull(['applicant2InRefuge'])
```
adamg-hmcts: I don't understand why we reset the refuge question specifically based on "applicant2ConfirmContactDetails"? Wouldn't the refuge answer be reset based on whether they've changed their answer to the confidentiality question, rather than whether they're happy with the contact details provided by their partner?

If we wanted to reset based on "applicant2ConfirmContactDetails", I'd have thought we'd reset all applicant 2 contact detail fields to get them to re-enter them, which would also solve the problem of the postcode search not showing for the address

FaisalMoJ: Removed the resetting field logic and will do it as post release enhancement as discussed

PR #4870, opened 2026-07-10
[NFDIV-5487] No Respondent Address Release Part 1

adamg-hmcts:
### Change description ###
https://tools.hmcts.net/jira/browse/NFDIV-5487

Early code release without turning on the new journey.


src/main/app/case/case-type.ts
```diff
@@ -1 +1 @@
-export const CASE_TYPE: string = process.env.CASE_TYPE || 'NFD';
+export const CASE_TYPE: string = 'NFD-5185';
```
FaisalMoJ, line 1: Do we need to revert this?

FaisalMoJ (approved):

